PR #10, opened 2015-04-05
Update osinfo.py with better interface

y12uc231:
I have added a new info about the release of the system along with a better looking result


geekcomputers:
Thanks

PR #16, opened 2015-10-22
added exceptions

festus64:
It would be a good idea to add exceptions to all scripts. You don't want your script to crash because of a simple error, do you?


geekcomputers:
Thank you for this, it’s a good idea

PR #22, opened 2015-12-15
Various Improvements

PythonProdigy:
1. Removed a redundant function call
2. Added exception handling to the file open function

### DIFF
--- a/check_file.py
+++ b/check_file.py
@@ -14,11 +14,14 @@ import os		# Import the Modules
 # Readfile Functions which open the file that is passed to the script
 
 def readfile(filename):
-	f = open(filename, 'r')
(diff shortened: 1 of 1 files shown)


check_file.py
```diff
@@ -9,29 +9,30 @@
...
+
+# Readfile Functions which open the file that is passed to the script
+
+def readfile(filename):
+	try:
+		f = open(filename, 'r')
+		line = f.read()
+		print line
+	except IOError:
+		print 'Error opening the file'
```
Angus3321: Here you forget to close the file after "read" it.Could we add a "finally" syntax ?And what if I read a file which is up to several G and may make it out of memory ?

PR #36, opened 2016-02-14
Remove unecessary line of code and variable, and tidy up comments and syntax.

jaredmanning:
I'm still learning Git, so sorry about the two different and unrelated commits in this.

The first one (70271c5) just tidies up the comments and syntax a bit in backup_automater_services.py. I added spaces on either side of some equal signs to make them easier to read.

The second one (0f47eb3) removes a line of code and a variable in check_file.py by combining the line 

    f = open(filename, 'r')

with the line below to make it

    line = open(filename, 'r').read()

Nothing crazy. Just some thoughts I had while checking these out. I'm currently learning Python and Git, so bear with me if I'm doing something wrong here.

I also did the same thing for check_for_sqlite_files.py in 903d620.

Thanks for considering this!

### DIFF
--- a/backup_automater_services.py
+++ b/backup_automater_services.py
@@ -1,30 +1,31 @@
 # Script Name	: backup_automater_services.py
 # Author			: Craig Richards
 # Created			: 24th October 2012
-# Last Modified	:
-# Version			: 1.0
+# Last Modified	: 13th February 2016
+# Version			: 1.0.1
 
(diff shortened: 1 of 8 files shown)


jaredmanning:
I think I went a little crazy on that... sorry if that is too much.

geekcomputers:
No worries, thanks for making the improvements

Cheers
Craig

PR #46, opened 2016-04-03
Add new file for the get info from remoute servers

teror4uks:
add new file for get info remoute servers


teror4uks:
I take to the basis this script https://gist.github.com/1284249 and rewrite input and output information for the list servers

PR #53, opened 2016-05-11
start  local server

ashwini0529:
This will start a local server on the mentioned port.
This is a good utility tool.

### DIFF
--- a/start-server.py
+++ b/start-server.py
@@ -1,0 +1,11 @@
+import SimpleHTTPServer
+import SocketServer
+
+PORT = 8000 #This will serve at port 8080 
+
+Handler = SimpleHTTPServer.SimpleHTTPRequestHandler
+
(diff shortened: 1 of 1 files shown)


ashwini0529:
Don't we need this?

ashwini0529:
@geekcomputers, please reply.

geekcomputers:
start local server? What do you mean

ashwini0529:
If you'll run this script, this will start a HTTP Client server on your localhost.
Run this script and open up your browser, and goto `localhost:8000` , you can see the directory listing.
This is very useful for understanding local servers.

geekcomputers:
Yes it a very handy fetaure, glad you like it :-)

ashwini0529:
I guess, that should be merged if there's no harm in doing that? ;)

PR #55, opened 2016-05-22
convert indent to space / reformat files based on PEP8

kyo-bad:
Tabs and space are mixed in some files.
So I convert indent to space and reformat files based on PEP8


geekcomputers:
Thanks

PR #76, opened 2016-09-18
Updates for CountMillionCharacter.py

walidmujahid:
Instead of having the string directly in the code as was in the original CountMillionCharacters.py, a person may load in a text file with their desired text for the code run with.


z843248880:
greate

PR #111, opened 2016-11-29
NameError: global name 'fd' is not defined

lyywo333:
File "check_for_sqlite_files.py", line 32, in <module>
    if isSQLite3(files):
  File "check_for_sqlite_files.py", line 22, in isSQLite3
    fd.close()
NameError: global name 'fd' is not defined


lyywo333:
NameError: global name 'fd' is not defined

PR #124, opened 2017-02-12
Added PEP 8 formatting and some minor fixes

yashLadha:


yashLadha:
please do merge as per the compatibility.

geekcomputers:

.idea/Python.iml
```diff
@@ -0,0 +1,12 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
EroMonsterSanji: I think you should use .gitignore to ignore these files.

PR #132, opened 2017-03-13
making the path in a safer way in create_dir_if_not_there.py

usrbit:
little improvements to the way the path is made up, used os.path.join()
and it's Python3 compatible


geekcomputers:

PR #142, opened 2017-03-28
Using String Formatting new format and list gens

ELC:


geekcomputers:

PR #145, opened 2017-04-03
Code to perform google search from terminal

ankit167:
The script is used to search for a keyword in google from the terminal. Top x (say 5) web pages will be displayed in separate tabs in the browser.


geekcomputers:

PR #146, opened 2017-04-03
Exception Handling in tweeter.py

ankit167:
Handling exceptions while calling twitter api to update status, with/without media.
f.e.: When character limit exceeds 140, the api call returns an exception.
      The exception is handled and the error is printed. This signifies an
      unsuccessful operation.


geekcomputers:

PR #149, opened 2017-04-06
Update new_script.py

OguzhanCtnkaya:


OguzhanCtnkaya:
wrong
print text 
print text
print 'Unknown option - ' + text

correct
print (text)
print (text)
print ('Unknown option - ' + text)

PR #166, opened 2017-05-16
Add files via upload

sahrulhs:


ex20.py
```diff
@@ -0,0 +1,33 @@
+from sys import argv
+
+script, input_file = argv
+
+def print_all(f):
+	print f.read()
+
```
ankit167: Please use two spaces between functions.

ex20.py
```diff
@@ -0,0 +1,33 @@
...
+current_file = open(input_file)
+
+print "First let's print the whole file:\n"
+print_all(current_file)
+
+print "Now let's rewind, kind of like a tape."
+rewind(current_file)
+
+print "Let's print three lines:"
+current_line = 1
```
ankit167: Better to use a loop. Would reduce Lines of Code.

ex20.py
```diff
@@ -0,0 +1,33 @@
...
+	print f.read()
+
+# seek(n) to read a file's content from byte-n
+def rewind(f):
+	f.seek(0)
+
+def print_a_line(line_count, f):
+	print line_count, f.readline()
+
+current_file = open(input_file)
```
ankit167: Please wrap non-function code inside main()

mam8: If the **input_file** is not existing, the **open** will fail.

ex20.py
sahrulhs: I type this code on debian's gedit, in gedit nothing automatic indentation, so the result like this...

ex20.py
sahrulhs: This code run on command line, so must there is a file in same directories.

mam8: I am also a newbie. I just learned a good way for this. It's with ... as ...: 
with open( input_file) as current_file:

ankit167: @sahrulhs The code is generic. Ideally, it should handle every case, which includes the file not being present at all.

PR #167, opened 2017-05-20
Computation for Factorial Permutation and Combination

ebiwari:
This code help in computation of factorial permutation and combination.. I will also update it to find Some more equation like quadratic, linear equation etc.

### DIFF
--- a/factorial_perm_comp.py
+++ b/factorial_perm_comp.py
@@ -1,0 +1,79 @@
+# Script Name		: factorial_perm_comp.py
+# Author			: Ebiwari Williams
+# Created			: 20th May 2017
+# Last Modified		: 
+# Version			: 1.0
+
+# Modifications		: 
+
+# Description		: Find Factorial, Permutation and Combination of a Number
+
+
+def factorial(n):	
+	fact = 1
+	while(n >= 1 ): 
+		fact = fact * n 
+		n = n - 1
+
+	return fact
+
+
+def permutation(n,r):
+	return factorial(n)/factorial(n-r)
+
+
+
+def combination(n,r):
+	return permutation(n,r)/factorial(r)
+
+
+def main():
+	print('choose between operator 1,2,3')
+	print('1) Factorial')
+	print('2) Permutation')
+	print('3) Combination')
+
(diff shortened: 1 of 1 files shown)


factorial_perm_comp.py
```diff
@@ -0,0 +1,79 @@
...
+# Version			: 1.0
+
+# Modifications		: 
+
+# Description		: Find Factorial, Permutation and Combination of a Number
+
+
+def factorial(n):	
+	fact = 1
+	while(n >= 1 ):
```
ankit167: extra whitespace before ')'

factorial_perm_comp.py
```diff
@@ -0,0 +1,79 @@
...
+def factorial(n):	
+	fact = 1
+	while(n >= 1 ): 
+		fact = fact * n 
+		n = n - 1
+
+	return fact
+
+
+def permutation(n,r):
```
ankit167: missing whitespace after ','

factorial_perm_comp.py
```diff
@@ -0,0 +1,79 @@
...
+	return factorial(n)/factorial(n-r)
+
+
+
```
ankit167: extra blank line

factorial_perm_comp.py
```diff
@@ -0,0 +1,79 @@
...
+def main():
+	print('choose between operator 1,2,3')
+	print('1) Factorial')
+	print('2) Permutation')
+	print('3) Combination')
+
+	operation = input('\n')
+
+	if(operation ==  '1'):
```
ankit167: extra space after operator '=='

factorial_perm_comp.py
ankit167: Base case can be written to check if r == 0. In this case, simply return 1 instead of calculating the factorial unnecessarily.

factorial_perm_comp.py
ankit167: It would be convenient for the user to enter 1,2 or 3 as input (as per the instructions), rather than '1','2' or '3'. Please check against integer values- 1,2 and 3

PR #169, opened 2017-05-27
Refactor code to uphold DRY principle.

nnamdiib:
The numerous "K.replace()" lines at the bottom can be refactored using a list of tuples.
A simple loop can then do all the replacing necessary.
And if you want to add more characters to replace, just update the list as required.

### DIFF
--- a/calculator.py
+++ b/calculator.py
@@ -45,12 +45,10 @@ def main():
     print ("\nScientific Calculator\nEg: pi * sin(90) - sqrt(81)")
 
     k = input("\nWhat is ")
-
-    k = k.replace(' ', '')
-    k = k.replace('^', '**')
-    k = k.replace('=', '')
-    k = k.replace('?', '')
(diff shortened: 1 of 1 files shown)


Angus3321:
Don't Repeat Yourself Principle

calculator.py
```diff
@@ -45,12 +45,10 @@ def calc(k):
...
-
-    k = k.replace(' ', '')
-    k = k.replace('^', '**')
-    k = k.replace('=', '')
-    k = k.replace('?', '')
-    k = k.replace('%', '/100')
+    replace_list = [(' ', ''), ('^', '**'), ('=', ''), ('?', ''), ('%', '/100')]
+    
+    for tup in replace_list:
+      k = k.replace(tup[0], tup[1])
```
cclauss:
```python
for old, new in replace_list:
    k = k.replace(old, new)
```

PR #172, opened 2017-05-29
NameError fix calculator.py

ayush1999:
using the normal input() function was causing a NameError.
changing it to raw_input() fixed this.

### DIFF
--- a/calculator.py
+++ b/calculator.py
@@ -44,7 +44,7 @@ def main():
 
     print ("\nScientific Calculator\nEg: pi * sin(90) - sqrt(81)")
 
-    k = input("\nWhat is ")
+    k = raw_input("\nWhat is ") # Using input() function is causing NameError. Changing it to raw_input() fixes this.
 
     k = k.replace(' ', '')
     k = k.replace('^', '**')


shreydan:
@ayush1999 this program is written in Pyrhon 3.

shreydan:
@ayush1999 also I do have a NameError Exception in the program and it works.

calculator.py
```diff
@@ -44,7 +44,7 @@ def calc(k):
 
     print ("\nScientific Calculator\nEg: pi * sin(90) - sqrt(81)")
 
-    k = input("\nWhat is ")
+    k = raw_input("\nWhat is ") # Using input() function is causing NameError. Changing it to raw_input() fixes this.
```
Angus3321: @ayush1999 python3 input will give you a string no matter what you input.py2 is different so,raw_input =>string,and input may result in a number if you input like"1+2".if wrong,thanks for telling me.

ayush1999: In python 2, raw_input will always return a string, however input will treat it as an expression.
In Python 3, input returns a string. Maybe this helps!

geekcomputers:

ayush1999:
Can someone please tell me how *args and **kwargs work in Python? 
I am having difficulty in using them.
Thanks!

EroMonsterSanji:
```*args``` is used to send a **non-keyworded** variable argument list to the function.
```**kwargs``` is used to pass **keyworded** variable of arguments to a function.

```python
def foo(first, *args, **kwargs):
    print('Required argument: ', first)
    for v in args:
        print('Optional argument (*args): ', v)
    for k, v in kwargs.items():
        print('Optional argument %s (*kwargs): %s' % (k, v))

>>> foo(1, 2, 3, 4, k1=5, k2=6)
```
Required argument:  1
Optional argument (*args):  2
Optional argument (*args):  3
Optional argument (*args):  4
Optional argument k2 (*kwargs): 6
Optional argument k1 (*kwargs): 5

PR #173, opened 2017-05-29
Latest Update_ for all varsions of interpreters

ayush1999:
I have added a new input_function. Based on the version of interpreter that the person is using , different functions will be called.
Since 2.7 DOES NOT support input() function, this bug has been fixed.
Now , we can use this calculator for all interpreters.

### DIFF
--- a/calculator.py
+++ b/calculator.py
@@ -1,10 +1,8 @@
 """
 Written by: Shreyas Daniel - github.com/shreydan
-Description: Uses Pythons eval() function 
+Description: Uses Pythons eval() function
              as a way to implement calculator
-
 Functions available:
-
 + : addition
 - : subtraction
 * : multiplication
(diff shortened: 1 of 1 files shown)


calculator.py
```diff
@@ -18,41 +16,46 @@
...
 
-    print ("\n" + str(calc(k)))
-    
+    print("\n" + str(calc(k)))
+
+
 if __name__ == "__main__":
-    main()
+
+    if sys.version_info.major< 3:
```
mam8: It seems the lines of 57 ~ 60 should be put into function of **main**.

calculator.py
```diff
@@ -18,41 +16,46 @@
...
+        input_function=raw_input
+    else:
+        input_function=input
+main()
```
mam8: The main() should be put in if __name__ == "__main__":

calculator.py
```diff
@@ -18,41 +16,46 @@
...
-        
-        return k
 
+        return k
 
-    print ("\nScientific Calculator\nEg: pi * sin(90) - sqrt(81)")
+    print("\nScientific Calculator\nEg: pi * sin(90) - sqrt(81)")
 
-    k = input("\nWhat is ")
+    k = input_function("\nWhat is ")
```
mam8: Could we change k to a more meaningful name, such as **raw_expression**?

calculator.py
```diff
@@ -18,41 +16,46 @@
...
+        functions = ['sin', 'cos', 'tan', 'sqrt', 'pi']
+
         for i in functions:
             if i in k.lower():
                 withmath = 'math.' + i
                 k = k.replace(i, withmath)
-        
+
         try:
             k = eval(k)
```
mam8: here, could we have the statement as **result** = eval(k)

geekcomputers:

PR #175, opened 2017-06-02
Add some judgement when changing sides.

Angus3321:


geekcomputers:

PR #176, opened 2017-06-04
more pythonic

ZivLi:


AkashSharma93:
Hi.

I think a better solution would be:
`print range(2, 100, 2)`

ZivLi:
@AkashSharma93 Awesome! But I think it may be more pythonic to use List Comprehensions , rather than just a method of Range to output some numbers.

geekcomputers:

PR #177, opened 2017-06-05
python3 compatibility for dir_test.py

EroMonsterSanji:
Use ```input()``` instead of ```raw_input()``` when the interpreter is 3.X,
and I put the functions into ```main()```


geekcomputers:

PR #178, opened 2017-06-06
Modified Check_file.py

EroMonsterSanji:
I reformatted this script, use '#' to separate files output. And moved the output part into the if block to avoid undefined 'filenames'


geekcomputers:

PR #180, opened 2017-06-07
Updated some comment and print lines.

SlimxShadyx:
I updated my comment lines as some users said that they were too long and expanded into multiple lines.
I have also broken one of the print lines into two, in case the same were to be happening with that as well.


geekcomputers:

PR #185, opened 2017-06-17
Update dice_rolling_simulator.py

CaptMcTavish:
Someone suggested me not to use while too much, so tried to create an alternative. 
I am a newbie and this is my first contribution!
A line in the previous code (The equivalent to which in the current code is line #26 said [10], which should have been [12], though its a minor slip, please fix it.) Thanks.

### DIFF
--- a/dice_rolling_simulator.py
+++ b/dice_rolling_simulator.py
@@ -1,98 +1,77 @@
 #Made on May 27th, 2017
 #Made by SlimxShadyx
+#Editted by CaptMcTavish, June 17th, 2017
 
 #Dice Rolling Simulator
 
 import random
 
-#These variables are used for user input and while loop checking.
-correct_word = False
-dice_checker = False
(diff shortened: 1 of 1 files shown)


dice_rolling_simulator.py
```diff
@@ -1,98 +1,77 @@
...
-
-        user_dice_chooser = int(user_dice_chooser)
-
-        if user_dice_chooser == 6:
-            dice_checker = True
+def result():
 
-        elif user_dice_chooser == 8:
-            dice_checker = True
+    user_dice_chooser
```
EroMonsterSanji: `user_dice_chooser` is referenced before assignment
Remove this line or comment it.

CaptMcTavish: Thanks, no idea how that got in there.

geekcomputers:

PR #186, opened 2017-06-28
Update README.md

SanthoshShanmugam:
Simply modified the sentence(First timer).


ytho:
Your 'fix' is incorrect. The original sentence was grammatically correct.

SanthoshShanmugam:
Thank you for the reply.

PR #187, opened 2017-07-02
All the 4 Digit Number Combinations

abrahamalbert18:
It'll print every 4 digit combination from 0000 to 9999. 
These are used in many real world applications.


abrahamalbert18:
Yeah sure. Thanks :) zfill() help my code look much better and will be useful.

PR #188, opened 2017-07-04
Update 4 Digit Number Combinations.py

ytho:
Removed repeated code and added PEP8 styling.


abrahamalbert18:
Thank you @ytho . I've updated the code using your suggestion. Appended the strings to the list, if the user needs to access it.

PR #195, opened 2017-07-11
Add line_count and character counting capability

sandynigs:
First contribution in open source world.
Write few lines of code , now it also count number of lines and total characters in the summoned file.


sandynigs:
[fileinfo.zip](https://github.com/geekcomputers/Python/files/1137990/fileinfo.zip)

Do have a look here.
Thank you.

Issue #200, opened 2017-07-14
Changing a list while iterating over it may cause unpredictable result.


fanofxiaofeng:
In check_file.py, the list `filenames` is changed on line 40 and line 47.
Thus, some items may skip the for loop (and does not get the existance check and read permission check), when you try to read such items on line 54, an error may be thrown.

For example, suppose there is non-readable file f1 and non-exist file f2 in current directory, and we execute the command `python check_file.py f1 f2`, we will see an error as shown in the picture.
![image](https://user-images.githubusercontent.com/3983683/28247458-6d237034-6a63-11e7-91b8-0f56cd9e0518.png)

geekcomputers:
Please feel free to change it on the repo

ArighnaIITG:
Is this issue closed?

fanofxiaofeng:
I forgot to close it.

PR #204, opened 2017-07-16
deal with issue 200

fanofxiaofeng:
Issue 200 is described here: [https://github.com/geekcomputers/Python/issues/200](url)
After modification, check_file.py can process non-exist files and non-readable files properly.
Suppose we have non-readable file f1 and non-exist file f2 in current directory, the result is shown in the picture.
image


fanofxiaofeng:
After modification, check_file.py can deal with non-exist files and non-readable files properly.

PR #206, opened 2017-07-20
convert decimal to hexadecimal

yadneshk:
converts any decimal value to hexadecimal value


geekcomputers:

Issue #209, opened 2017-07-24
dice.py remove 'self' from self.sides_change


vishesh1042:
sides_change parameter is passed to the function set_sides. Within the function set_sides, instead of using the sides_change variable directly, self.sides_change is used. It is resulting in an error.

made2591:
Fix issue #209 adding more significant print messages and removing self.sides_change in method.

PR #210, opened 2017-07-27
Fix issue #209

made2591:
Fix issue #209 adding more significant print messages and removing self.sides_change in method.


geekcomputers:

PR #231, opened 2017-09-16
fix to show all the matches

Digres45:
- Added variable 'l' instead of fix constant number 10


geekcomputers:

PR #238, opened 2017-09-23
A simple program illustrating chaotic behaviour

tanakachingonzo:
This program also demonstrates the apparent unpredictability of chaos functions and how small initial differences can lead to large differences as the formulae is repeatedly applied.


geekcomputers:

PR #243, opened 2017-10-02
Keep image extension for xkcd comics

watersalesman:
It might be best to just keep the file extension, to prevent confusion. Especially on Windows where how the file is opened depends on the file extension.


geekcomputers:

PR #245, opened 2017-10-03
minor fix to function oneLineCombinations()

NaeemSheikh:
I ran the second function and it only cycled from 0000 to 0999 as the range in the program is range(1000).
Doesn't really need fixing, as people will understand that the range in the code needs an extra 0, but I would like to submit this change of range to range(10000) just for the sake of completion.


geekcomputers:

PR #247, opened 2017-10-08
Update multiplication_table.py

Starix36:
If we are going to use the higher version of python like 3.3.4 than we might get a possible error in the print statements if open and close parentheses are not applied.


geekcomputers:

PR #261, opened 2017-11-17
Update and rename myip.py to other_pepole/get_ip_gui

khaled15486:
i try to convert it to GUI , for practice :)


geekcomputers:

PR #263, opened 2017-11-22
Re-made the dice.py code into a new dice-program. 

DevDoggo:
This program is more dynamic in the sense that you can change sides and amounts of dices in real-time. You can specify up to an very, very large (max value of the python integer) amount of sides and dices, however you do have a certain lower-limit to sides and amounts of dices to 'realistically'
represent the dices in a 3-dimensional space. I've used several generalized functions, rather than several similar special-cases, to minimize code use and I hope people may get some inspiration from it.

The name possibly looks terrible, so I'm more than willing to re-consider it.

### DIFF
--- a/diceV2_dynamic.py
+++ b/diceV2_dynamic.py
@@ -12,7 +12,7 @@ class Dice():
         if sides > 3:
             self.sides = sides
         else:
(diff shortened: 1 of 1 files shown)


geekcomputers:

PR #267, opened 2017-12-04
Added check if file present

mcostigliola:
If the file is present it continues in the execution of the program, otherwise it manages an exception and asks to try again


geekcomputers:

Issue #274, opened 2017-12-16
Facing issues with the youtube script.


pjha1994:
~/Desktop/GIT/other/Python$ python youtube.py 
Enter the song to be played: hero
Traceback (most recent call last):
  File "youtube.py", line 28, in <module>
    song = songs[0].contents[0].contents[0].contents[0]
IndexError: list index out of range

sidharthshah:
I think you're facing this issue either because they keyword "hero" couldn't be found. Try putting in some alternative keyword{s} and see if it still works

cclauss:
@pjha1994 Can this issue be closed?

PR #283, opened 2018-01-05
removed headers from youtube.py

rtatu:
Headers are removed, the user agent is for Macintosh and not compatible with other operating system so better to be removed for issue #274


geekcomputers:

PR #286, opened 2018-01-13
Print message when sides set to 6

hemantkarandikar:
Hello!
I am newbie to Python. I am new to Github also. So please pradon me if I don't follow some procedures or make a mistake. Thanks.

Here is why I am suggesting a change:
When d.set_sides(6) there is no way of knowing that the dice has six sides. This change is bringing uniformity.


geekcomputers:

PR #289, opened 2018-01-16
Update Organise.py

leojacob7:
Added a change by which if there is a file like filename.py.txt then the system will convert to folders folder,even though it is a document.
Using the last element in the list i.e name.split('.')[-1] it will get the last element alone.
Also added comments. properly


geekcomputers:

PR #297, opened 2018-02-05
Update factors.py

sinamohammadi99:
a must be an int number and input() take a string char so we have to use int()


geekcomputers:

PR #298, opened 2018-02-07
Async downloading example

dspetrenko:
It's example of usage asyncio+aiohttp to downloading.

Dependencies:
- Python 3 (3.5) 
- aiohttp (2.1.0)

Start the example:

You start the example with python async_downloader.py in the console.
In test case script try download web pages to files.


geekcomputers:

PR #309, opened 2018-03-15
Added a calculate PI function

MonliH:
I added a calculate pi function. I used the Chudnovsky algorithm

### DIFF
--- a/primelib/primelib.py
+++ b/primelib/primelib.py
@@ -38,6 +38,25 @@ fib (n) // calculate the n-th fibonacci term.
 goldbach(number)  // Goldbach's assumption
 
 """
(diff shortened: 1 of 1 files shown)


primelib/primelib.py
```diff
@@ -602,4 +621,4 @@ def fib(n):
         ans += fib1
         fib1 = tmp
         
-    return ans
```
MonliH: This change was accidental.

primelib/primelib.py
```diff
@@ -602,4 +621,4 @@ def fib(n):
...
\ No newline at end of file
+    return ans
```
MonliH: After I created this pull request I noticed this. This change was accidental.

PR #330, opened 2018-04-16
Updated by adding XML Parser

mohitsharmaaa:
Previous contributor added only a url that was not needed.
Added LXML Parser that is needed to parse XML page.


geekcomputers:

PR #336, opened 2018-05-22
Calculator recovers after errors & strengthen comment language.

jhbiggs:
Hi, I changed the exception functions so that the calculator can keep going even if it encounters an error.  I also strengthened some the language in some of the comments.


geekcomputers:
Thanks

PR #344, opened 2018-06-19
Update CountMillionCharacter.py

Colinjog:
Better use dict, dict can delete the duplicate words.


geekcomputers:

PR #350, opened 2018-07-03
add unionpay card

cyjstifler:
add china Unionpay card


cyjstifler:
add china Unionpay card

geekcomputers:

PR #375, opened 2018-10-02
Update and rename Shivaansh.txt to Shivaansh.py

dareghost:
added some content 
a program to print a table up 10 of any given number


geekcomputers:

PR #376, opened 2018-10-02
Update ReadFromCSV.py

dareghost:
corrected the modules which were imported 
and imported numpy which was required for the given code


geekcomputers:

PR #377, opened 2018-10-02
Update calculator.py

dareghost:
did error handling for type error when the user gives an input of incorrect datatype and added the exponents function


geekcomputers:

PR #385, opened 2018-10-05
Removing #!/usr/bin/python3 from Google Image Downloader

Mosrod:
I removed #!/usr/bin/python3 from 3 files


geekcomputers:

cclauss:
Why is it a good idea to remove [shebang lines](https://en.wikipedia.org/wiki/Shebang_(Unix))?

PR #390, opened 2018-10-07
hacktober fest

jathinjohny:
iam a beginner


geekcomputers:
Why have you changed that line, that’s the variable name from above?

PR #394, opened 2018-10-08
Removed system argument and removed the first row multiplication table

nraikm:
Fixed the code of the multiplication table so it works without the system argument using an input, and removed the first row when it prints, so it's only 1 row - argv rows.

### DIFF
--- a/multiplication_table.py
+++ b/multiplication_table.py
@@ -1,14 +1,9 @@
-from sys import argv # import argment variable 
+argv = input('Number of ROWS and COLUMBS: ')
 
-script, rows, columns = argv # define rows and columns for the table and assign them to the argument variable
+script, rows, columns = argv, argv, argv # define rows and columns for the table and assign them to the argument variable
 
 
 def table(rows, columns):
-	for i in range(1, int(rows) + 1 ): # it's safe to assume that the user would mean 12 rows when they provide 12 as an argument, b'coz 12 will produce 11 rows
-		print ("\t", i,)
-
-	print ("\n\n") # add 3 lines
-
 	for i in range(1, int(columns) + 1 ):
 		print (i),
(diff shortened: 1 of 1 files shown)


multiplication_table.py
```diff
@@ -1,14 +1,9 @@
-from sys import argv # import argment variable 
+argv = input('Number of ROWS and COLUMBS: ')
 
-script, rows, columns = argv # define rows and columns for the table and assign them to the argument variable
+script, rows, columns = argv, argv, argv # define rows and columns for the table and assign them to the argument variable
```
cclauss: The __input()__ statement is asking for two values, not three.  We should also deal gracefully with commas and spaces in the input: __rows, columns = argv.strip().replace(",", "").split()__

multiplication_table.py
```diff
@@ -1,14 +1,9 @@
...
 
 
 def table(rows, columns):
-	for i in range(1, int(rows) + 1 ): # it's safe to assume that the user would mean 12 rows when they provide 12 as an argument, b'coz 12 will produce 11 rows
-		print ("\t", i,)
-
-	print ("\n\n") # add 3 lines
-
 	for i in range(1, int(columns) + 1 ):
 		print (i),
```
cclauss: __print (i),__ --> __print(i, end="")__ # for Python 3

cclauss: Also put __from \_\_future\_\_ import print_function__ at the top of the file and use the line __print("\t",i*j, end="")__ near the bottom of the file.

multiplication_table.py
```diff
@@ -1,14 +1,9 @@
-from sys import argv # import argment variable 
+argv = input('Number of ROWS and COLUMBS: ')
```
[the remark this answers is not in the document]
paulwang007: 👍

PR #395, opened 2018-10-08
Added in to the file to fix error in Print_List_of_Even_Numbers.py

Mosrod:
Fixed the error in line 6, by adding 'in' in between 'i' and 'range' in Print_List_of_Even_Numbers.py


geekcomputers:

PR #396, opened 2018-10-08
Prints list of odd numbers

Mosrod:
There was a script for even numbers, so I decided to add a script for odd numbers.


geekcomputers:

PR #399, opened 2018-10-08
Updated python code

Aniruddha09:
Changed Single-line import to multi-line imports according to [PEP-8](https://www.python.org/dev/peps/pep-0008/#imports)


geekcomputers:

PR #417, opened 2018-10-19
Added Decimal to Binary Converter

Corruption13:
I've added a complete Decimal (Base 10) Number to Binary Number converter to the root folder.


geekcomputers:

PR #419, opened 2018-10-26
Added PSNR calculator for checking compression quality

vasutomar:
PSNR is used to calculate how well the compression algorithm works, it takes 2 images as input and outputs the PSNR value.


geekcomputers:

PR #420, opened 2018-10-26
Add weather scrapper to get details of current weather

mansiag:
The script takes city name as input and stores the current weather information of that city in csv file


geekcomputers:

PR #425, opened 2018-10-30
Update Google_News.py

devtycoon:
Added comment to function and updated print feature using the new format syntax.


geekcomputers:

PR #430, opened 2018-11-14
performance optimization of primelib.isPrime(number)

WinterWonderland:
only loop over the odd numbers.

### old version:

In [1]: from primelib import isPrime

In [2]: %timeit isPrime(99999999999999997)
50.9 s ± 31.1 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

### new version:

In [1]: from primelib import isPrime

In [2]: %timeit isPrime(99999999999999997)
32.4 s ± 149 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

### DIFF
--- a/primelib/primelib.py
+++ b/primelib/primelib.py
@@ -73,15 +73,15 @@ def isPrime(number):
     
     # 0 and 1 are none primes. 
     if number <= 1:
-        status = False
-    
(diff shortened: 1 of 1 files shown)


geekcomputers:

PR #439, opened 2018-12-03
Update to python3 format

jinyoungHan:
Change python2 from file to python3 format


jinyoungHan:
While testing, I had a Python version issue and changed some files to python3.

Is it okay if I try to change the rest of the files?

PR #444, opened 2018-12-14
Update to Python 3 format

jinyoungHan:
I Updated to Python 3 format

### DIFF
--- a/PORT SCANNER.PY
+++ b/PORT SCANNER.PY
@@ -78,9 +78,9 @@ remoteServer    = raw_input("Enter a remote host to scan: ")
 remoteServerIP  = socket.gethostbyname(remoteServer)
 
 # Print a nice banner with information on which host we are about to scan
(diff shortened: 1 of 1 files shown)


PORT SCANNER.PY
```diff
@@ -116,5 +116,5 @@ t2 = datetime.now()
 total =  t2 - t1
 
 # Printing the information to screen
-print 'Scanning Completed in: ', total
+print('Scanning Completed in: ', total)
```
cclauss: __print__ statements __that contain a comma__ need special treatment on Python 2.  
* __python2 -c "print('a', 'b')"__  # --> ('a', 'b')
* __python2 -c "from __future__ import print_function ; print('a', 'b')"__  # --> a b

jinyoungHan: Thanks for your help

PR #457, opened 2019-01-08
Fix undefined name 'filename' in ftp_send_receive.py

cclauss:
Also uses __with open() as__ syntax to ensure that file handles are properly closed.
```
./ftp_send_receive.py:24:27: F821 undefined name 'filename'
	ftp.retrbinary('RETR ' + filename, LocalFile.write, 1024)
                          ^
```
@gepcel  Your review please.

### DIFF
--- a/ftp_send_receive.py
+++ b/ftp_send_receive.py
@@ -18,7 +18,7 @@ ftp.cwd('/Enter the directory here/')
 	Enter the location of the file where the file is received
 """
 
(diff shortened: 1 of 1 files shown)


gepcel:
This is better, more pythonic. If we are gonna follow the pep8, it will be better to use lowercase function names, like `receive_file`  and `send_file`.  But I'm fine with `ReceiveFile` too.

PR #547, opened 2019-10-02
Update EncryptionTool.py

abhis2007:
Line no 55 & 56 can be cancatenate using list comprehension method


abhis2007:
Line of code can be reduced by at least 1 line. I create some changes in line 55&56 .

PR #568, opened 2019-10-03
Update Print_List_of_Even_Numbers.py

shivagrwl312:


Print_List_of_Even_Numbers.py
```diff
@@ -18,5 +18,8 @@ def print_error_messages(): #function to print error message if user enters a ne
   if(n==0):
     print(start) #prints 0 first as it is an even number
     start+=2 #increases the value of start by 2 to get to the next even number
+    patch-1
+  elif(n%2==0):
   else(n%2==0):
+    master
     print(start)
```
QuantumNovice: ^ This is not the way of doing it in python
```python
for i in range(0, start, 2):
   print(i)
```

PR #569, opened 2019-10-03
Update Print_List_of_Odd_Numbers.py

shivagrwl312:


Print_List_of_Odd_Numbers.py
```diff
@@ -18,6 +18,9 @@ def print_error_messages():
   if(n==0):
     start+=1
     print(start)
-  else:
-    if(n%2!=0):
-    print(start)
+    patch-2
```
NitkarshChourasia, line 21: patch is not defined anywhere else.
I must say.
So, it will raise error. Because it is not mapped/declared/initiated.

Print_List_of_Odd_Numbers.py
```diff
@@ -18,6 +18,9 @@ def print_error_messages():
...
+  elif(n%2!=0):
+    else:
+      if(n%2!=0):
+        master
```
NitkarshChourasia, line 25: master is not defined anywhere else.
I must say.
So, it will raise error. Because it is not mapped/declared/initiated.

Print_List_of_Odd_Numbers.py
```diff
@@ -18,6 +18,9 @@ def print_error_messages():
   if(n==0):
     start+=1
     print(start)
-  else:
-    if(n%2!=0):
-    print(start)
+    patch-2
+  elif(n%2!=0):
```
NitkarshChourasia, line 22: If proper coding formatting was done, it would have been better.

Print_List_of_Odd_Numbers.py
```diff
@@ -18,6 +18,9 @@ def print_error_messages():
...
+    else:
+      if(n%2!=0):
+        master
+      print(start)
```
NitkarshChourasia, line 26: Lack of proper documentation makes it hard to understand the purpose of the program to others.
It would have been better if it was properly documented.

Print_List_of_Odd_Numbers.py
NitkarshChourasia, line 21: If possible use a function.
**WHY?**
That is the first step to good programming practices.
Function, that is the basic to writing to good programs.
Implement it.

NitkarshChourasia (changes requested):

PR #598, opened 2019-10-12
Enforce an interactive backend

smiley3:
The script currently runs with the default matplotlib backend. However, it needs an interactive backend.

On my machine, the default happened to be AGG which is not an interactive backend and resulted in the following error.

`game_o_life.py:117: UserWarning: Matplotlib is currently using agg, which is a non-GUI backend, so cannot show the figure.
  fig.show()`

It now relies on TkAGG which use tk that should be included by default.

### DIFF
--- a/game_of_life/game_o_life.py
+++ b/game_of_life/game_o_life.py
@@ -31,6 +31,9 @@ import random
 import sys
 
 import numpy as np
+
+from matplotlib import use as mpluse
+mpluse('TkAgg')
 from matplotlib import pyplot as plt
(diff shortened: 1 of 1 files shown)


game_of_life/game_o_life.py
```diff
@@ -31,6 +31,9 @@
 import sys
 
 import numpy as np
+
+from matplotlib import use as mpluse
+mpluse('TkAgg')
```
smiley3: Note that this should be called before importing pyplot. Otherwise, it has no effect.

PR #602, opened 2019-10-13
Update Print_List_of_Odd_Numbers.py

Ayush144:
chnged somewthing


Ayush144:
chnged print-odd-number.py

PR #608, opened 2019-10-16
importing turtle

babugyan:


wolf99:
Maintainers, please check requester's profile. Has only commits  all on the same period, all making similar spurious commits. No idea why but I can't think of any good reasons.

PR #620, opened 2019-10-18
Added another way to compute

subahanii:
for pelindrome and factorial


subahanii:
> Why again int(n) which is already done typecasting taking input above lines
Yes ,You are right.
now removed.

subahanii:
> What is complexity of [::-1]

of course O(n)

PR #675, opened 2019-11-28
removing extra space

ajay1472:


ajay1472:
removing extra space from script

PR #705, opened 2020-03-23
format the string & optimize counting

Alopex4:
Use `{}` as placeholder to format string could make the code look more clear.
The `file_stats` is iterable so use `zip` with string list can generate a directory.

The former counting line/chart isn't concise enough, it could be replace by `sum`

### DIFF
--- a/fileinfo.py
+++ b/fileinfo.py
@@ -1,4 +1,4 @@
- Script Name       : fileinfo.py
+# Script Name       : fileinfo.py
 # Author            : Not sure where I got this from
 # Created           : 28th November 2011
 # Last Modified     : 23th March 2020
--- a/fileinfo.py
+++ b/fileinfo.py
@@ -1,12 +1,11 @@
 # Script Name       : fileinfo.py
-# Author                : Not sure where I got this from
-# Created               : 28th November 2011
-# Last Modified     :
-# Version               : 1.0
+# Author            : Not sure where I got this from
+# Created           : 28th November 2011
+# Last Modified     : 23th March 2020
+# Version           : 1.0
 # Modifications     :
 
-# Description           : Show file information for a given file
-
+# Description       : Show file information for a given file
 
 # get file information using os.stat()
 # tested with Python24 vegsaeat 25sep2006
@@ -16,24 +15,31 @@ import os
 import stat  # index constants for os.stat()
 import sys
 import time
+from collections import OrderedDict
 
 if sys.version_info >= (3, 0):
     raw_input = input
 
-file_name = raw_input("Enter a file name: ")  # pick a file you have
+# pick a file you have
+file_name = raw_input("Enter a file name: ")  
 count = 0
 t_char = 0
+
 try:
     with open(file_name) as f:
-        line = f.readline()
-        t_char += len(line)
-        while line:
-            count += 1
-            line = f.readline()
-            t_char += len(line)
+        # Source: https://stackoverflow.com/a/1019572
+        count = (sum(1 for line in f))
+        f.seek(0)
+        t_char = (sum([len(line) for line in f]))
 except FileNotFoundError as e:
     print(e)
-    sys.exit()
+    sys.exit(1)
+# When open item is a directory (python2)
+except IOError:
+    pass
+# When open item is a directory (python3)
+except IsADirectoryError:
+    pass
 
 file_stats = os.stat(file_name)
 # create a dictionary to hold file info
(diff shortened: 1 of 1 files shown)


fileinfo.py
```diff
@@ -50,24 +56,31 @@
...
+                     "st_ino (inode number),"
+                     "st_dev (device),"
+                     "st_nlink (number of hard links),"
+                     "st_uid (user ID of owner),"
+                     "st_gid (group ID of owner),"
+                     "st_size (file size bytes),"
+                     "st_atime (last access time seconds since epoch),"
+                     "st_mtime (last modification time),"
+                     "st_ctime (time of creation Windows)")
+    stats_keys = stats_strings.split(',')
```
MichaelCurrin: How about creating stats_strings and a list then there is no need to split?

fileinfo.py
```diff
@@ -50,24 +56,31 @@
...
+    for key, value in OrderedDict(zip(stats_keys, file_stats)).items():
```
MichaelCurrin: You can leave out OrderedDict and .items, as iterating through key and value of the zipped pairs will be sufficient.

fileinfo.py
```diff
@@ -50,24 +56,31 @@
...
 }
 
-print("\nfile name =", file_info['fname'])
-print("file size =", file_info['fsize'], "bytes")
-print("last modified =", file_info['f_lm'])
-print("last accessed =", file_info['f_la'])
-print("creation time =", file_info['f_ct'])
-print("Total number of lines are =", file_info['no_of_lines'])
-print("Total number of characters are =", file_info['t_char'])
+file_info_fmt = ('\nfile name = {}\nfile size = {}\nlast modified = {}\n'
```
MichaelCurrin: This has become a lot less readable by separating keys and values. And by having last accessed and creation time on the same line.

fileinfo.py
```diff
@@ -50,24 +56,31 @@
...
-    print("A closer look at the os.stat(%s) tuple:" % file_name)
-    print(file_stats)
-    print("\nThe above tuple has the following sequence:   ")
-    print("""st_mode (protection bits), st_ino (inode number), 
-    st_dev (device),    st_nlink (number of hard links),    
-    st_uid (user ID of owner),   st_gid (group ID of owner),    
-    st_size (file size, bytes),  st_atime (last access time, seconds since epoch),  
-    st_mtime (last modification time),   st_ctime (time of creation, Windows)"""
-          )
+    file_stats_fmt = ''
```
MichaelCurrin: Remove if not used.

Alopex4: You are right, I forget to delete this line.

fileinfo.py
Alopex4: Great! It could make the code more concise, thanks your opinion.

Alopex4: Great! It could make the code more concise, thanks your opinion.

fileinfo.py
Alopex4: Agree, it looks more nice. Sometime I just too lazy to create a list(contain string) , so I would like to use `string with commas` as raw data, and then `split` it to list. It's a good choice or not ? Can you give me an answer?

MichaelCurrin: No, that's a bad choice. It makes the code less readable and adds an extra unnecessary step to read and to compute.

Also you just have if you move the comma one character from `,"` to `",` for each line (which is no change in laziness), plus you get to leave off the `.split` line` so its actually less code.

Also, it is error prone - what if there are commas in the middle of the string that you _don't_ want to split on?

fileinfo.py
MichaelCurrin: Cool, thanks for updating.

fileinfo.py
```diff
@@ -23,17 +22,22 @@
...
-            t_char += len(line)
+        # Source: https://stackoverflow.com/a/1019572
+        count = (sum(1 for line in f))
+        f.seek(0)
+        t_char = (sum([len(line) for line in f]))
 except FileNotFoundError as e:
     print(e)
-    sys.exit()
+    sys.exit(1)
+# When open item is a directory (python2)
```
cclauss: Do we still need to support legacy Python?  https://pythonclock.org/

fileinfo.py
Alopex4: Oh you remind me that. Actually in the previous version because the `','` inside the string, I find the print out result can't satisitify my expect. You help me a lot, thank you so much. :+1:

fileinfo.py
Alopex4: Ha ha, actually because those code was write in  28th November 2011. So may be it should be keep the code still support previous python verison.

PR #710, opened 2020-04-09
Program that Replaces Specified Text String In a Text File

LinjianLi:


MichaelCurrin:
For interest, this can be done in a one line `sed` command 

    sed 's/foo/bar/g' hello.txt

[Link](https://github.com/MichaelCurrin/cheatsheets/blob/master/cheatsheets/shell/substitute_text.md)

But the Python approach is nice as part of a larger application.

LinjianLi:
> For interest, this can be done in a one line `sed` command
> 
> ```
> sed 's/foo/bar/g' hello.txt
> ```
> 
> [Link](https://github.com/MichaelCurrin/cheatsheets/blob/master/cheatsheets/shell/substitute_text.md)
> 
> But the Python approach is nice as part of a larger application.

Fair point. I was just being consistent with the repo name "Python" and this script is good for Windows users and Linux newbies like me (lol).

PR #736, opened 2020-05-19
Printing list of even numbers using the list methods

abhinavanand1608:


Print_List_of_Even_Numbers.py
```diff
@@ -1,9 +1,11 @@
-# INPUT NUMBER OF EVEN NUMBERS
+n = int(input("Enter the required range : "))  # user input
+list = []
 
-n = int(input("Enter the limit : "))  # user input
+if (n < 0):
+    print("Not a valid number, please enter a positive number!")
+else:
+    for i in range(0,n+1):
```
Moshiurmilton, line 7: 0 is not an even number so there should be range(1, n+1)

PR #798, opened 2020-08-29
Update JARVIS.py

SteveParadox:


SteveParadox:
specifying language to translate and giving it the ability to show all words heard.

PR #812, opened 2020-09-15
Create number of a particular letter in a message

divu050704:
want to have a program to find a particular letter in a syntax with a input from the user


NebilT (commented):
M

PR #834, opened 2020-09-25
Update digital_clock.py

gabesduhbear:
When I ran your code, I got an error on line 24 which stated "NameError: name 'clack' is not defined." Because of this, I tried changing the word clack to the function dig_clock and then it worked! To learn more about the arguments that tkinter's after function takes, read this https://www.geeksforgeeks.org/python-after-method-in-tkinter/. Have fun and good job!


aayush6200:
is this issue solved yet?? It not I would be happy to solve it

PR #865, opened 2020-10-01
Update binod.py

anand722000:
Comment always helps us to understand code in fast. I am also a beginner, like us, many friends are there, it will help full for them.
Thank you...


anand722000:
Currently, I am learning how GitHub works it will be thankful if you will help me. thank you

PR #870, opened 2020-10-01
Update stack.py

Swastik-Saha:
Added peek() function and improved isEmpty() function.

### DIFF
--- a/stack.py
+++ b/stack.py
@@ -11,9 +11,8 @@ def size(stack):
 	return len(stack) 
 
 # Stack is empty if the size is 0 
-def isEmpty(stack): 
-	if size(stack) == 0: 
-		return True 
+def isEmpty(stack):
+	return stack == [] 
 
 # Function to add an item to stack . 
 # It increases size by 1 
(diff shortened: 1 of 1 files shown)


stack.py
```diff
@@ -46,7 +45,15 @@ def reverse(string):
...
+# A function to see the top of stack without removing it
+def peek(stack):
+    if isEmpty(stack):
+        return "Error : Underflow"
+    else:
+        top = len(stack) - 1
+        return stack[top]
 	
 # Driver program to test above functions 
 string="GeeksQuiz"
```
bupboi1337, line 52: You could replace string and with just str to save a inch of space

stack.py
```diff
@@ -46,7 +45,15 @@ def reverse(string):
 	for i in range(0,n,1): 
 		string+=pop(stack) 
 		
-	return string 
+	return string
+
+# A function to see the top of stack without removing it
+def peek(stack):
+    if isEmpty(stack):
+        return "Error : Underflow"
```
bupboi1337, line 53: Maybe tell the user to restart the program and report the bug

bupboi1337 (commented):
Doesn't look broken to me but further looking into is required.

PR #897, opened 2020-10-01
friday

aasthasaxena217:
Created a small program like Jarvis.


meezan-mallick:
Your program fails in the lint_python check.

There is an improper indentation at line no 13
`print("anything more")`
remove the extra tab, Hope this will work.

PR #938, opened 2020-10-02
Greatone

soni-111:
program to find the greatone of two numbers in python


implosion07 (commented):
try to put your code in a try except block so that in case of a string entry no traceback is detected

PR #966, opened 2020-10-02
Update README.md

Deepak-dash007:
Updated Content in Readme - Hactober


unnown:
Please mark as invalid and spam. 
This is not what hacktoberfest is about and people like this should be banned

PR #980, opened 2020-10-03
Update create_dir.py

ishanisamanta:
Edited comments


Roshaen (commented):
Not a major improvement

PR #984, opened 2020-10-03
Create patternprint

vignesh-sura:
this program is used to print pattern


vignesh-sura:
accept for hacktoberfest and this code is done by me.

PR #996, opened 2020-10-04
Update README.md

Sh1710:


majormoses:
This is likely spam, brand new user recently created and then a bunch of updates to readmes that don't really make much sense, following the same patterns.s. See https://blog.domenic.me/hacktoberfest/, sad state of oss when maintainers have to deal with this over some free t-shirts.

PR #1011, opened 2020-10-05
sum of even factors of a number

heyamrita:
Given a number n, the task is to find the even factor sum of a number.


heyamrita:
umm

PR #1020, opened 2020-10-05
Create convert celsius into fahrenheit

pankaj1278:
using this code we can easily convert   celsius into fahrenheit


pankaj1278:
hlo

PR #1051, opened 2020-10-06
Added the ability for a user to chose if they want the file created or not

jcovington16:
Just wanted to add some more interaction with the script. Thought of a use case where a user might not want the directory created or possibly could've mistyped the name of the directory in question. I probably could shorten this more, but for now it works just fine.


jcovington16:
I'm confused. Are there packages that I need to run with this? I ran the file and everything worked just fine. Could I get more insight as of what failed?

PR #1064, opened 2020-10-06
Create how to add three numbers and find type in python

Mansi-1120:
simple program


Mansi-1120:
Please add level hactober fest accepted

PR #1068, opened 2020-10-06
find Armstrong number

ayshmishra688:
program to find Armstrong number


SwapnilBais:
The easiest way to find the Armstrong number 
[Animal_crackers.pdf](https://github.com/geekcomputers/Python/files/5334231/Animal_crackers.pdf)

PR #1102, opened 2020-10-08
Added nasa_apod_with_requests(py) which can get you the Astronomy Photo of the Day of given date using Requests.

Lunaditu:


Lunaditu:
My commit doesn't complete the lint check because of some other files on the repo...

PR #1128, opened 2020-10-14
swap numbers

Ak-12343994967:
To swap numbers


ganesh-thorat01:
No need to add temp variable just do
x, y=y,x
And it will be swapped.

PR #1139, opened 2020-10-15
Update whatsapp-monitor.py

anandshaurya011:
"Pip3" cammond dosen't works in all os
Here's a best cammond python -m pip install.....


anandshaurya011:
Any queries...?

DrStarkXavier:
@anandshaurya011 

you've to install the pip packages
 then you can use it. you can do that by running this command
`sudo apt-get install pip`

And then you can run this package to install any other packages related to *python3*

For instance,
`python3 -m pip install #package_name`

PR #1152, opened 2020-10-16
Python Program to Remove Punctuations from a String

Mayank-Singla9:
Python Program to Remove Punctuations from a String


Tony-NanJing:
You can use the input() function.

PR #1181, opened 2020-10-19
Create to Check whether Year is a Leap Year or not

akhilesh147:
In this program, user is asked to enter a year. Program checks whether the entered year is leap year or not.


akhilesh147:
In this post, we will write a Python program to check whether the input year(entered by user) is leap year or not.

PR #1217, opened 2020-11-01
Create PythonToHash

Anon007-cyber:


PythonToHash
```diff
@@ -0,0 +1,27 @@
+# Python rogram to find the SHA-1 message digest of a file
```
kashifkhan0771, line 1: Python program to find the SHA-1 message digest of a file( you missed a word `p` in program
And keep this comment below imports for better readability

kashifkhan0771 (approved):

PythonToHash
```diff
@@ -0,0 +1,27 @@
...
+
+# importing the hashlib module
+import hashlib
```
bupboi1337, line 4: If your importing modules always remember to keep a requirments.txt listing the modules imported in the same folder as the program itself.

PythonToHash
```diff
@@ -0,0 +1,27 @@
...
+       chunk = 0
+       while chunk != b'':
+           # read only 1024 bytes at a time
+           chunk = file.read(1024)
+           h.update(chunk)
+
+   # return the hex representation of digest
+   return h.hexdigest()
+
+message = hash_file("track1.mp3")
```
bupboi1337, line 26: could replace message with just m and comment that m means message for smaller filesize.

PythonToHash
bupboi1337, line 1: typo but good description of the program overall with this single comment.

bupboi1337 (changes requested):
### Its good but needs some changes
This looks like a wonderful script but it does have some issues that can be fixed.
All it needs is the following for it to check out in my book are the following things:

- The program to be in its own folder
-  A requirments.txt listing the module(s) used (in this case it would be the hashlib module)
- And last but not least some minor optimization and rewording

After all that is added this would be a great addition to add to this repo.
**tl;dr it looks great but needs a bit more work**

NitkarshChourasia (approved):
Properly documented.
Properly named.
Done, following the best practices and standards of python programming language, indeed.

PR #1240, opened 2020-11-21
New algorithim

erdum:
First time contribution

### DIFF
--- a/Electronics_Algorithms/Ohms_law.py
+++ b/Electronics_Algorithms/Ohms_law.py
@@ -1,0 +1,12 @@
+def ohms_law(v=0, i=0, r=0):
+    if v == 0:
+        result = (i*r)
+        return result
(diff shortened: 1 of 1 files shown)


erdum:
What is the problem i did not understand there is no issue with my file but with someone's else, so what i can do for to pass checks????

Lugixion:
That's a problem with other files so don't worry about it.

Electronics_Algorithms/Ohms_law.py
```diff
@@ -0,0 +1,12 @@
+def ohms_law(v=0, i=0, r=0):
```
bupboi1337, line 1: Add comments for more readability

bupboi1337 (commented):
All this needs is some more comments.

bupboi1337:
man, 1 year

PR #1243, opened 2020-11-22
Update Sum of digits of a number

ftwood:
Single-line version of the solution


ftwood:
So, my solution has no bugs and errors)

PR #1260, opened 2020-12-28
Changed from index method to if method

georgerahul24:
Changed  the method from index to if and else so as to make simplify the code


georgerahul24:
Can anyone please explain why it failed the check..
the file for which it failed the check was not modified by me

georgerahul24:
It yet failed the check again

PR #1267, opened 2021-01-02
QR Code Generator with tkinter GUI

GauravSingh9356:


SimpleCalculator.py
```diff
@@ -17,6 +21,7 @@ def divide(a, b):
 def power(a,b):
     return a**b
 
+
 def main():
```
bupboi1337, line 25: Add more comments/documentation so more people understand your code.

SimpleCalculator.py
```diff
@@ -1,4 +1,8 @@
+<<<<<<< HEAD
```
bupboi1337, line 1: What do these even do?!

SimpleCalculator.py
```diff
@@ -45,4 +50,6 @@ def main():
     else:
         print("Invalid input")
         main()
+
```
bupboi1337, line 53: why are these here too?

bupboi1337 (changes requested):
Good idea / concept. Not the best implementation.

PR #1275, opened 2021-01-22
Update main.py

VesselKreative:
As i was trying to run the script it raised the error.
Traceback (most recent call last):
  File "c:/Users/zedthecode/Desktop/zedlab/gui/tube.py", line 13, in <module>
    if a == "shutdown":
NameError: name 'a' is not defined


jakubwrz:
You need to define "a" first... but why is this if-statement two times there?

PR #1299, opened 2021-03-08
Add Dani555555 to Contributors list

Dani555555:
choose a number and this program will give you the list of divisors


implosion07 (commented):
seems good

PR #1304, opened 2021-03-19
Fixed the data conversion failure

LUHUIBIN:


dji77:
Fix the data conversion

dji77 (commented):
Is doing something

PR #1313, opened 2021-04-16
I changed the email id-dictionary program. 

gedelanavya:
Instead of using two loops, I used max and count function


gedelanavya:
I didn't even open that wave_sort.py file.

PR #1315, opened 2021-04-20
Loading message in python

shaurya-blip:


shaurya-blip:
What is wrong with this?

loader.py
```diff
@@ -0,0 +1,38 @@
+"""
+Shaurya Pratap Singh 
+@shaurya-blip
+
+Shows loading message while doing something.
+"""
+
+import itertools
```
bupboi1337, line 8: Always list dependencies in a requirements.txt file!

loader.py
```diff
@@ -0,0 +1,38 @@
...
+import threading
+import time
+import sys
+
+# The task is not done right now
+done = False
+
+
+def animate(message="loading", endmessage="Done!"):
```
bupboi1337, line 17: For smaller file size I would replace message with start and endmessage with end

bupboi1337 (approved):
Seems like it would be a amazing addition to this repo.

bupboi1337:
> What is wrong with this?

Lint-python fails when it thinks it finds mistakes in your code. I would check out the log for it and fix what caused it to fail.

shaurya-blip:
Found anything yet?

PR #1335, opened 2021-06-10
Update Binary_search.py

tanujdhiman:
Iterative approach take time. And time complexity is high. Instead of using this approach we can use recursive approach. It take less time as compare to Iterative approach.

Thanks


slowy07 (approved):

tanujdhiman:
Hello @slowy07 !!

Thanks for approving changes !!

PR #1339, opened 2021-06-29
Update calc_area.py

RedPlumDragon:
Adds 3 more (3d) shapes, instead of just multiplying by 3.14 I changed it to pi, and I also added an option to run the program again.

### DIFF
--- a/calc_area.py
+++ b/calc_area.py
@@ -1,19 +1,42 @@
 # Author: PrajaktaSathe
 # Program to calculate the area of - square, rectangle, circle, and triangle -
-shape = int(input("Enter 1 for square, 2 for rectangle, 3 for circle, or 4 for triangle: "))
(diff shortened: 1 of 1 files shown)


calc_area.py
```diff
@@ -1,19 +1,42 @@
...
-elif shape == 4:
-  base = float(input("Enter base: "))
-  h = float(input("Enter height: "))
-  print("Area of rectangle = " + str(0.5*base*h))
-else:
-  print("You have selected wrong choice.")
\ No newline at end of file
+import math as m
+
+def main():
```
kashifkhan0771, line 5: use a descriptive name for this function like `calculate` or `cal_area`. `main` is too simple

kashifkhan0771 (approved):
Feedback

PR #1392, opened 2021-09-30
3 different functions to check a prime number

henry-the-vietnamese:
I define **three** different functions (with very different algorithms) to check if a number is prime.
Those three functions, from a to c, decrease in efficiency (takes longer time).
With three different functions able to compute one result, I hope that students can benefit from grasping the idea behind them.

### DIFF
--- a/check_prime.py
+++ b/check_prime.py
@@ -1,0 +1,41 @@
+# Author:   Tan Duc Mai
+# Email:    tanducwork@gmail.com
+
+"""
+Three different functions to check whether a given number is a prime.
+Return True if it is a prime, False otherwise.
+Those three functions, from a to c, decreases in efficiency (takes longer time).
+"""
+
+
+def is_prime_a(n):
+    from math import sqrt
+    if n < 2:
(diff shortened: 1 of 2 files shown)


check_prime.py
```diff
@@ -0,0 +1,41 @@
...
+
+"""
+Three different functions to check whether a given number is a prime.
+Return True if it is a prime, False otherwise.
+Those three functions, from a to c, decreases in efficiency (takes longer time).
+"""
+
+
+def is_prime_a(n):
+    from math import sqrt
```
slowy07, line 12: should on the top

slowy07 (changes requested):
LGTM, some cnages modules

check_prime.py
henry-the-vietnamese: That's a good idea. I'll fix it.
Thanks.

henry-the-vietnamese:
> LGTM, some cnages modules

Done. Thanks for your suggestion.

PR #1397, opened 2021-10-01
Update print_colors.py

omijagtap:
Description for colors.py

### DIFF
--- a/Colors/print_colors.py
+++ b/Colors/print_colors.py
@@ -10,7 +10,7 @@ class colors:
 
 def printc(color, message):
     print(color + message + colors.ENDC)
-
+#color which we print or import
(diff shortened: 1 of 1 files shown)


Colors/print_colors.py
```diff
@@ -10,7 +10,7 @@ class colors:
 
 def printc(color, message):
     print(color + message + colors.ENDC)
-
+#color which we print or import
```
slowy07, line 13:
```suggestion
# color which we print or import
```

slowy07 (changes requested):

Colors/print_colors.py
omijagtap: I have make a changes

PR #1398, opened 2021-10-01
Update README.md

digvesh301:


Assembler/README.md
```diff
@@ -1,5 +1,5 @@
 # Python-Assembler
-# WE NEED A FREE T-SHIRT
+# WE need A FREE T-SHIRT
```
henry-the-vietnamese, line 2: I'd say leave it as it was since lower case only one word results in _inconsistency_. And the change made does not make any sense.

henry-the-vietnamese (changes requested):

PR #1409, opened 2021-10-21
Added Emoji dictionary

akash-rajak:
Added emoji dictionary in python.


Emoji Dictionary/requirements.txt
```diff
@@ -0,0 +1,3 @@
+libraries used : tkinter
+		 messagebox
+		 emoji
```
slowy07, line 3: just type the module, for example

```
tkinter
messagebox
emoji
```
so you can install by typing on cmd or terminal by
```
pip install -r requirements.txt
```

PR #1410, opened 2021-10-21
Update Binary_search.py

Rohanrbharadwaj:
Function name was mentioned wrong while recurring, so lint_python was failing


bupboi1337 (approved):
This looks good and should be merged.

PR #1424, opened 2021-11-06
Security updates and more requirments.txt files.

bupboi1337:
Updated vulnerable libraries to the latest versions and added some more requirments.txt files.


PDF/requirements.txt
```diff
@@ -1,2 +1,2 @@
-Pillow==5.0.0
+Pillow==8.3.2
```
bupboi1337, line 1: Might cause some incompatibilities but it would fix the security bugs

async_downloader/requirements.txt
```diff
@@ -1 +1 @@
-aiohttp==2.1.0
+aiohttp==3.7.4
```
bupboi1337, line 1: might also be a bit broken but will fix security bugs too

bupboi1337 (commented):
this looks like it would work

PR #1465, opened 2022-01-19
Update question 5.py

GOTAM672:
closing the file


implosion07 (commented):
Try to make code more compact and short using methods applicable on string

PR #1469, opened 2022-01-27
Fix big security bug

bupboi1337:
this bug is dangerous and we should patch it


bupboi1337 (commented):
all this does is it just updates a library with some bugs.

DontEatThemCookies:
Does this relate to CVE-2022-22815 through 22817?

bupboi1337:
i guess

DontEatThemCookies:
I see

DontEatThemCookies (approved):

PR #1504, opened 2022-04-24
Create Phishing app

mhnasajpour:


DontEatThemCookies (changes requested):
I don't recommend pushing this, as this kind of code that can easily be misused for malicious purposes. At the very least, a disclaimer in the comments such as "Strictly for educational purposes only" could be inserted, but that still can't stop bad actors from maliciously using it. I'd rather not push this at all.

slowy07:
yeah same to me sir @DontEatThemCookies, i think it become useless, and look that, its seem have some useless function and ddnt have explanation by commenting for each function

shaischaudhry (approved):

mhnasajpour:
May you merge this?

PR #1526, opened 2022-06-07
Removed unnecessary function(utils.validte)

cyai:
Removed the unnecessary function, i.e. utils.validate, as it was not necessary and the input was already validated in the previous loop!


cyai:
I am so sorry to open the same pull request on two different commits, that was unintentional and I am not finding a way to revert it! 
Sorry for the inconvenience!

PR #1557, opened 2022-06-21
Rename HangMan Game to HangMan-Game.py

cyai:


slowy07:
it think the old name be better than new name, and good for python naming convention

PR #1561, opened 2022-06-22
Improved wiki/wiki.py

Cerabbite:
I found that there were some isseus with the wiki.py file. I found that once you got your result of the first query you could enter any query you wanted but the output would not change. Also sometimes it would not show any output at all.

I fixed isseu one by putting everything in a main class and then declaring answer in the __init__ instead of in the summary.
The second issue I could not fix, but I used tkinters messagebox.showerror to indicate to the user when something is wrong.

Then I also binded the enter key to running the summary function so instead of heaving to click the search button you can also press enter. When you search the text on the search button changes to searching to indicate the program is searching.


shaischaudhry (approved):

slowy07 (approved):

Cerabbite:
Last 2 commits are not for this pull request

PR #1612, opened 2022-07-16
Linitng issue for flake-8 fixed

Bakhtaki:
It was an issue with Flake -8 which is fixed now!!!

### DIFF
--- a/cartesian_product.py
+++ b/cartesian_product.py
@@ -1,7 +1,6 @@
 """Cartesian Product of Two Lists."""
 
 # Import
-from itertools import product
 
 
 # Cartesian Product of Two Lists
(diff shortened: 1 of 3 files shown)


Bakhtaki:
can some one help with the below issue:
flake8 . --count --select=E9,F63,F7,F82 --show-source --statistics

I have checked my code with flake-8 several times and there is no issue in my editor, but during the pull request check, it will fail with the mentioned error code.

Bakhtaki:
> can some one help with the below issue: flake8 . --count --select=E9,F63,F7,F82 --show-source --statistics
> 
> I have checked my code with flake-8 several times and there is no issue in my editor, but during the pull request check, it will fail with the mentioned error code.

It is done, it was an issue with another file, and I was checking just my own files, now it is Done, and all checks passed utterly.

PR #1662, opened 2022-08-10
Rename python_webscraper to python_webscraper.py

cyai:


slowy07 (commented):
this pull not change some code

PR #1678, opened 2022-08-22
Application Fix

swe-pea:
Fixed outdated/broken application, new update implements Selenium to acheive a working solution.


swe-pea:
Any clues on how to gain approval from contributors?

PR #1679, opened 2022-08-22
Example of generators

LuisZapataYamo:
Example of generators using fibonacci numbers

### DIFF
--- a/FibonacciNumbersWithGenerators.py
+++ b/FibonacciNumbersWithGenerators.py
@@ -1,0 +1,18 @@
+def fibonacciGenerator(n = None):
+    """
+        Generating function up to n fibonacci numbers iteratively
+        Params:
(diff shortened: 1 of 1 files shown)


FibonacciNumbersWithGenerators.py
```diff
@@ -0,0 +1,18 @@
+def fibonacciGenerator(n = None):
```
cclauss, line 1: Python uses snake_case for function names.
```suggestion
def fibonacci_generator(n = None):
```

LuisZapataYamo: Thanks for the sugestions

cclauss (approved):

LuisZapataYamo (commented):
Add example of generators

PR #1698, opened 2022-09-27
Update BruteForce.py

kalilinux2009:
updated and serialised some variable values


rein14:
Could uou give more insight on why dis this

PR #1699, opened 2022-09-29
Refactored and Optimised Code base

rahulgurujala:
Optimized all code bases and refactored code.


BlackJack_game/blackjack.py
OfficialAhmed, line 1: `sum(d_cards)` is not evaluated. It should be evaluated and 
Sum of both `d_cards` & `p_cards` should equal to `21`

convert_time.py
OfficialAhmed, line 1: * Removing comments not a good idea as its not enhancing the code in any way unless 
* One way to factor this block of code is to assign `input_str[-2]` to a variable, this way its more readable hence the comments are no longer needed 👍

coronacases.py
OfficialAhmed, line 1: `match-case` is faster than `if-else`. If you could do that instead of the if-else
<div> <code>match choice:
    case "1":
        ...
    case _:
        ...
</code> </div>

OfficialAhmed (approved):

OfficialAhmed (approved):

OfficialAhmed (approved):

convert_time.py
NitkarshChourasia: yes.

BlackJack_game/blackjack.py
NitkarshChourasia: yes.

coronacases.py
NitkarshChourasia: yes, @OfficialAhmed is right.
Match case are being introduced in not but recent version of python.
They are somewhat new in to this language.

CRC/crc.py
```diff
@@ -49,7 +47,7 @@ def crc_check(data, div):
     print("Receiver side remainder: ", remainder)
     print("Continue [Y/N]:")
     ch = input()
-    if ch == "N" or ch == "n":
+    if ch in ["N", "n"]:
         break
     else:
         continue
```
danila12315454, line 53: can be deleted

PR #1790, opened 2022-11-12
Create quo.md

gerrishons:
quo frame work for creating clis


Sd-Shiivam (approved):
I think there is an issue with the primary workflow.

PR #1873, opened 2023-03-25
Add Handling Exceptions..

muzakkirsaifi123:
I have updated the code with handling exceptions. Now the user is able to add or input their own path. So users have more flexibility.


muzakkirsaifi123:
I have checked the code but I do not know why it is showing a syntax error even though I have not changed the Fibonacci code but it is showing. Please have a look and merge my code. If there is any mistake in the GitHub action then i also like to check it.

PR #1875, opened 2023-03-27
Update basic.py

nikitapandeyy:
I noticed that the comment on the line where the font size is set to 18 says 28 instead of 18.

### DIFF
--- a/PDF/basic.py
+++ b/PDF/basic.py
@@ -1,34 +1,20 @@
 from fpdf import FPDF
 
-# Author: @NavonilDas
-
-
 pdf = FPDF()
-# Set Author Name of the PDF
 pdf.set_author("@NavonilDas")
-# Set Subject of The PDF
 pdf.set_subject("python")
-# Set the Title of the PDF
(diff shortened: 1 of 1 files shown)


PDF/basic.py
```diff
@@ -1,34 +1,20 @@
 from fpdf import FPDF
 
-# Author: @NavonilDas
```
NitkarshChourasia, line 3: Only this is what to be removed.
Even as optional.
But, as we are approaching a standardized open sourced, so removing this may be legitimate.
But, removing the comments which make the program better is not good.
Comments and documentations are the essence as they help in understanding the program much more better.

NitkarshChourasia (changes requested):

NitkarshChourasia:
## Comments and documentations are the essence as they help in understanding the program much more better.

### Don't remove them.
# Restore them, please.

### This is not the standard and best practices of programming and that too, python programming language.

lighting9999 (approved):

PR #1941, opened 2023-07-25
Update Swap.py using best practices and standards of python.

NitkarshChourasia:
The improvements made to the program are significant and can be summarized as follows:

 Object-Oriented Approach: The original code was a simple script with variables and direct swapping of values. In contrast, the improved version uses an object-oriented approach with a class Swapper that encapsulates the swapping logic and functionality.

 Class Methods: The swapping methods (swap_tuple_unpacking, swap_temp_variable, and swap_arithmetic_operations) are defined as class methods in the Swapper class. This allows them to be called on instances of the class and operate on the specific values of x and y associated with each instance.

Error Handling: The __init__ method in the improved version includes error handling to check if both x and y are integers. If not, a ValueError is raised, providing a clear message to the user about the expected input.

Encapsulation: The display of values before and after swapping is encapsulated within the display_values method. This method allows for consistent and easy printing of values without repeating the print statements in each swapping method.

Docstrings: The improved version includes detailed docstrings for the class and each method. Docstrings provide clear documentation about the purpose and usage of the class and its methods, making it easier for other developers (including the original programmer) to understand and use the code.

Examples with Class Methods: The improved version demonstrates the usage of class methods with two examples (swapper1 and swapper2). Each example creates an instance of Swapper with different initial values and then calls one of the swapping methods on the instance.

Reusability: With the improved version, the Swapper class can be easily reused in other parts of the code or in different programs, promoting code modularity and reusability.

Overall, the improved version is more organized, follows object-oriented principles, and provides better readability and maintainability compared to the original script-like code. It also incorporates error handling and documentation, making it more robust and user-friendly.

### DIFF
--- a/swap.py
+++ b/swap.py
@@ -1,12 +1,79 @@
-x = 5
-y = 10
+class Swapper:
+    """
+    A class to perform swapping of two values.
 
-# To take inputs from the user
-# x = input('Enter value of x: ')
-# y = input('Enter value of y: ')
+    Methods:
+    -------
+    swap_tuple_unpacking(self):
+        Swaps the values of x and y using a tuple unpacking method.
+    
+    swap_temp_variable(self):
+        Swaps the values of x and y using a temporary variable.
+    
+    swap_arithmetic_operations(self):
+        Swaps the values of x and y using arithmetic operations.
 
-# Swap the values of x and y without the use of any temporary value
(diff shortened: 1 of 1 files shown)


NitkarshChourasia:
Admin of this page, don't you like well written programs?

PR #1943, opened 2023-07-26
CODE OPTIMIZATION

OfficialAhmed:
* Implemented `match-case` for enhanced performance instead of `if-else` -> Faster by 86% ( [match-case benchmark](https://github.com/tonybaloney/anti-patterns/blob/master/bench_match.py#L3-L23) )
* Consolidating checks into a dynamic function, `setZeroFlag`:
  Combine multiple checks into a single dynamic function, `setZeroFlag` which takes `token` and `tmpToken` as arguments and returns a `bool` to set the zero flag

### DIFF
--- a/Assembler/assembler.py
+++ b/Assembler/assembler.py
@@ -70,847 +70,868 @@ def scanner(string):
 
     for ch in string:
 
-        if state == 0:
+        match state:
 
-            if ch == "m":  # catch mov-command
+            case 0:
+            
+                match ch:
 
-                state = 1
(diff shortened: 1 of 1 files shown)


NitkarshChourasia (approved):
Wonderfully used match case to make the code more readable and standardized.
Thus, the code is more properly formatted.

NitkarshChourasia:
Good, use of if and else cases.

NitkarshChourasia:
instead of it you used match case, good.

PR #1946, opened 2023-07-29
Update Armstrong_number.py

Piombacciaio:
Script can now work with all the numbers of lenght n and base 10

### DIFF
--- a/Add two numbers.py
+++ b/Add two numbers.py
@@ -7,4 +7,4 @@ num2 = float(input("Number 2:"))
 sum = num1 + num2
 
 # Display the sum
-print("The sum of"num1,'and'num2,'is',sum)
+print(f"The sum of {num1} and {num2} is {sum}")
--- a/Armstrong_number.py
+++ b/Armstrong_number.py
@@ -1,21 +1,24 @@
-def is_armstrong_number(number):
-    total = 0
+"""
+In number theory, a narcissistic number (also known as a pluperfect digital invariant (PPDI), an Armstrong number (after Michael F. Armstrong) or a plus perfect number), 
+in a given number base b, is a number that is the total of its own digits each raised to the power of the number of digits.
+Source: https://en.wikipedia.org/wiki/Narcissistic_number
+NOTE:
+this scripts only works for number in base 10
+"""
 
-    # find the sum of the cube of each digit
-    temp = number
-    while temp > 0:
-        digit = temp % 10
-        total += digit ** 3
-        temp //= 10
+def is_armstrong_number(number:str):
+    total:int = 0
+    exp:int = len(number) #get the number of digits, this will determinate the exponent
+
+    digits:list[int] = []
+    for digit in number: digits.append(int(digit)) #get the single digits
+    for x in digits: total += x ** exp #get the power of each digit and sum it to the total
     
-    # return the result
-    if number == total:
-        return True
+    # display the result
+    if int(number) == total:
+       print(number,"is an Armstrong number")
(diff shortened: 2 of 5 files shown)


OfficialAhmed (approved):

FIND FACTORIAL OF A NUMBER.py
```diff
@@ -3,7 +3,7 @@
 def factorial(n):
    if n < 0:
       return("Oops!Factorial Not Possible")
-   elif n = 0:
+elif n == 0:
```
OfficialAhmed, line 6: Double check the `indentation` here

Piombacciaio:
I'm slowly updating every file so that the lint check gets completed with no errors 
Had to stop for the moment, will resume (and hopefully finish) in the next 24 h

Armstrong_number.py
```diff
@@ -1,21 +1,24 @@
...
+this scripts only works for number in base 10
+"""
 
-    # find the sum of the cube of each digit
-    temp = number
-    while temp > 0:
-        digit = temp % 10
-        total += digit ** 3
-        temp //= 10
+def is_armstrong_number(number:str):
```
OfficialAhmed, line 9: The function could be slower than the older implementation. Due to the `for-loops`. I rather go with `1 while-loop` than `2 for-loops`. It would be much faster if you approach the problem with list comprehensions and the `sum()`

```python

total:int = sum( [int(digit) ** exp for digit in number] )

# display the result
...
...
...
```

OfficialAhmed (changes requested):
## Merging this request is not recommended yet

Add two numbers.py
```diff
@@ -7,4 +7,4 @@
 sum = num1 + num2
 
 # Display the sum
-print("The sum of"num1,'and'num2,'is',sum)
```
NitkarshChourasia, line 10: Good string formatting.

Armstrong_number.py
```diff
@@ -1,21 +1,24 @@
-def is_armstrong_number(number):
-    total = 0
+"""
```
NitkarshChourasia, line 1: ADded a docstring.

Armstrong_number.py
NitkarshChourasia: So while loop is faster then for loop?

Armstrong_number.py
```diff
@@ -1,21 +1,24 @@
...
-    # find the sum of the cube of each digit
-    temp = number
-    while temp > 0:
-        digit = temp % 10
-        total += digit ** 3
-        temp //= 10
+def is_armstrong_number(number:str):
+    total:int = 0
+    exp:int = len(number) #get the number of digits, this will determinate the exponent
+
```
NitkarshChourasia, line 12: Made program a little better to read and added some string formatting, good.

FIND FACTORIAL OF A NUMBER.py
NitkarshChourasia, line 6: Yes, double check it.

FibonacciNumbersWithGenerators.py
```diff
@@ -14,5 +14,5 @@ def fibonacci_generator(n = None):
         f0, f1 = f1, fn
         n -= 1
 
-for n_fibo in fibonacci(7):
+for n_fibo in fibonacci_generator(7):
```
NitkarshChourasia, line 17: good.

NitkarshChourasia (approved):
Yes, made program a little readable.

PR #1951, opened 2023-07-30
Development

NitkarshChourasia:
## Fixed Formatting in most of the programs.
# Made the program much more readable.


NitkarshChourasia:
@OfficialAhmed Please help me out in this resolve thing.

NitkarshChourasia:
@OfficialAhmed I tried solving it, solved it and even pushed it. Still it has the same program in merge conflicts.

NitkarshChourasia (commented):
All are perfectly done. Well done. Nitkarsh Chourasia.

OfficialAhmed:
Where is the problem ? 
Its better if you commit a small number of changes at a time, so its easier to solve.

OfficialAhmed (approved):

BoardGame-CLI/uno.py
```diff
@@ -99,9 +100,10 @@ def canPlay(colour, value, playerHand):
 numPlayers = int(input("How many players?"))
 while numPlayers < 2 or numPlayers > 4:
     numPlayers = int(
-        input("Invalid. Please enter a number between 2-4.\nHow many players?"))
+        input("Invalid. Please enter a number between 2-4.\nHow many players?")
+    )
 for player in range(numPlayers):
-    players_name.append(input("Enter player {} name: ".format(player+1)))
+    players_name.append(input("Enter player {} name: ".format(player + 1)))
```
OfficialAhmed, line 106: f-string is easier to read in this case

OfficialAhmed (approved):

Crack_password.py
```diff
@@ -1,11 +1,39 @@
 from random import *
+
 user_pass = input("Enter your password: ")
-password = ['a', 'b', 'c', 'd', 'e', 'f', 'g', 'h', 'i', 'j','k', 'l', 'm', 'n', 'o', 'p', 'q', 'r', 's', 't','u','v','w', 'x', 'y', 'z',]
+password = [
```
OfficialAhmed, line 4: For better performance a `set` or a `tuple` much quicker when reading password elements

OfficialAhmed (approved):

BoardGame-CLI/uno.py
NitkarshChourasia: Okay. Readability ❤️ is the 🔑

OfficialAhmed: F-string is 

'''python 
f"hello, {user}"
'''

No need to use `.format()`

Hotel-Management.py
```diff
@@ -39,73 +18,51 @@ def menu():
...
 
-    with open("Management.txt", "r") as File:
-        string = File.read()
-        string = string.replace("'", '"')
-        dictionary = json.loads(string)
+    with open("Management.txt", "r") as file:
+        dictionary = json.load(file)
 
-    if len(dictionary.get("Room")) == 0:
+    if not dictionary["Room"]:
```
OfficialAhmed, line 56: Its not recommended in this case to directly fetch "Room".. instead its a good practice to use `.get()` because if the value Room is not available in the dictionary the script will error and halt however `.get()` is going to return `None` type so the script can continue forward without errors

OfficialAhmed (approved):

FibonacciNumbersWithGenerators.py
```diff
@@ -1,18 +1,31 @@
...
+    Generating function for up to n Fibonacci numbers iteratively.
+    Params:
+        n: int
+    Return:
+        int
     """
     f0, f1 = 0, 1
     yield f1
-    while n == None or n > 1:
+    while n is None or n > 1:
```
OfficialAhmed, line 21: checking `None` is unnecessary for more readable code I'd something like 
```python
while not n or n>1
```
or perhaps 
```python
# if n is None loop terminates auto.
while n>1
```

FibonacciNumbersWithGenerators.py
```diff
@@ -1,18 +1,31 @@
...
         fn = f0 + f1
         yield fn
         f0, f1 = f1, fn
-        n -= 1
+        if n is not None:
```
OfficialAhmed, line 25: Easier to read
```python
if n:
```

OfficialAhmed (approved):

Patterns/pattern6.py
```diff
@@ -1,17 +1,18 @@
...
-#B B 
-#C C C 
-#D D D D 
-#E E E E E   
+# A
+# B B
+# C C C
+# D D D D
+# E E E E E
 def alphabetpattern(n):
```
OfficialAhmed, line 7: The Snake formatting is commonly used in `Python` not an issue but one way to make it easier for Python Devs
`Snake format` utilizes underscores to link the words
```python
def alphabet_pattern():
```

Python Program to Display Fibonacci Sequence Using Recursion.py
```diff
@@ -1,15 +1,16 @@
 def recur_fibo(n):
-   if n <= 1:
-       return n
-   else:
-       return(recur_fibo(n-1) + recur_fibo(n-2))
+    if n <= 1:
+        return n
+    else:
```
OfficialAhmed, line 4: `else` can be ignored in this case since we have a `return-statement` if its `True` then it'll exit the function, otherwise it'll 100% read the next statement. So another way to make it easier as follows:
```python
if n <=1:
    return n
return recur_fibo(n - 1) + recur_fibo(n - 2)
```

Python Program to Find LCM.py
```diff
@@ -1,20 +1,21 @@
 # Python Program to find the L.C.M. of two input number
 
+
 def compute_lcm(x, y):
+    # choose the greater number
+    if x > y:
```
OfficialAhmed, line 6: In this block I would ignore the `else` statement by initializing the variable with a default value first then just use an `if-statement` as follows:

```python
greater = y
if x > y:
    greater = x
```
its ideal to choose the default value which in most cases would be true like 

```python
n = input("Enter you name: ")
err = False
if not n:
    # if n is empty
    err = True
``` 

I assumed the input is going to be true by saying `err = False`
because in most cases the `n` is going to be entered by the user, there is no way the user can bypass to the next statement without entering the input unless its empty, therefore, `err=False`

OfficialAhmed:
BTW the failing check is not from your side

OfficialAhmed (approved):

OfficialAhmed (approved):
Most of  the commits aren't necessary, they aren't making the code any better in terms of the performance or readability nor user-friendly you might be ignored when trying to commit such edits to the main project for example:

```python
import os, json
```
You changed it to 

```python
import os
import json
```
another example the quotation marks

```python 
'hello'
``` 
to
```python 
"hello"
``` 
Python accepts both for a string so its unnecessary.

another reason 

```python
data = [
    [1][2],
    [3][4]
]
``` 
you changed it to 

```python
data = [[1][2], [3][4]]
``` 
both are correct but doesn't change the performance whatsoever. More like this type of commits for all 315 files is not ideal, since you didn't introduce nor implement a better way to enhance neither the code performance nor code readability. Good thing none of them has conflicts. Great job overall you went through all the files and changed few lines in each, however it depends now on the owner of this repo, if they wanted all these changes to be committed on the main project

NitkarshChourasia:
> Most of the commits aren't necessary, they aren't making the code any better in terms of the performance or readability nor user-friendly you might be ignored when trying to commit such edits to the main project for example:
> 
> ```python
> import os, json
> ```
> 
> You changed it to
> 
> ```python
> import os
> import json
> ```
> 
> another example the quotation marks
> 
> ```python
> 'hello'
> ```
> 
> to
> 
> ```python
> "hello"
> ```
> 
> Python accepts both for a string so its unnecessary.
> 
> another reason
> 
> ```python
> data = [
>     [1][2],
>     [3][4]
> ]
> ```
> 
> you changed it to
> 
> ```python
> data = [[1][2], [3][4]]
> ```
> 
> both are correct but doesn't change the performance whatsoever. More like this type of commits for all 315 files is not ideal, since you didn't introduce nor implement a better way to enhance neither the code performance nor code readability. Good thing none of them has conflicts. Great job overall you went through all the files and changed few lines in each, however it depends now on the owner of this repo, if they wanted all these changes to be committed on the main project

# INDEED @OfficialAhmed 

### But, It was all about following and adhering to the best standards and practices of Python programming language using PEP 8 as the best reference guide for the rules to be adhered to.

Thank you, for you valuable insight on the changes, made.

It seems like the owner of this repository doesn't seem to accept changes in multiple files rather then few files in the moment as a pull request.

NitkarshChourasia:
## @geekcomputers It seems like the owner of this repository doesn't seem to accept changes in multiple files all at ONCE rather then Few Files in the moment as a Pull Request.

sarayusreeyadavpadala:
Nice effort NitkarshChourasia.

OfficialAhmed (approved):

PR #1955, opened 2023-08-01
CODE REFACTORIZATION

OfficialAhmed:
CHANGED: 2 files missing `.py` extension

MERGED   : `pattern1.py` and `pattern3.py` since they're implemented in a similar way. 

The user can choose the pattern between:

- `repeat` which repeats the pattern such as `111 22 3` 
- `incremental` increments like `1 12 123`

The user can choose the pattern position:

- `upper` part
```python
1
12
123
```
- `lower` part
```python
123
12
1
```


Patterns/pattern1.py
NitkarshChourasia, line 1: Why this was deleted @OfficialAhmed . Was it repeating the same thing as the above Patterns/half triangle pattern.py?

Patterns/pattern3.py
NitkarshChourasia, line 1: Why this was deleted @OfficialAhmed . Was it repeating the same thing as the above Patterns/half triangle pattern.py?

Patterns/half triangle pattern.py
NitkarshChourasia, line 1: Wouldn't it be better if the name "half_triangle_pattern.py" instead of "half triangle pattern.py", the one in the beginning uses snake_case which adheres more to the best practices and standards of python programming language.

Calendar (GUI).py
NitkarshChourasia, line 1: I want your take @OfficialAhmed on naming file using snake_case as it is what is used in terms of python to name objects usually. So does naming files with that name adheres to the best practices and standards of python programming language? especially the files with extension ".py"

NitkarshChourasia (changes requested):

PR #1963, opened 2023-08-11
Comments are added

sarayusreeyadavpadala:
Hey geekcomputers! I have added some comments and made an indentation correction. Suggest if any changes required.


sarayusreeyadavpadala:
It is showing 1 failing check. I don't understand what's wrong with the code. I have executed it, got output correctly. Could you please help me.

end
```diff
@@ -0,0 +1,530 @@
+[33mcommit 6bb102f0bbf6c41ac1b766d8ee1c08e615262859[m[33m ([m[1;36mHEAD -> [m[1;32msum[m[33m, [m[1;31morigin/sum[m[33m)[m
```
NitkarshChourasia: It would be better, if this thing(file) is removed.

Sum of digits of a number.py
```diff
@@ -1,7 +1,13 @@
-q=0
-n=int(input("Enter Number: "))
-while(n>0):
- r=n%10
- q=q+r
- n=n//10
+q=0   # Initially we assigned 0 to "q", to use this variable for the summation purpose below.
```
NitkarshChourasia, line 1: You can improve it further.
- By taking number as a string.
- Checking for its numerical value.
- If good, then using split or loop function.
- And at the same time adding them.

This program lacks, 
Error management.
In real world suppose I gave it string input, then it would just crash, right?
What if I gave it a float number? Because float numbers has ' . ' (dot) in them, then again it would crash.
So, being pragmatic is the approach.

Sum of digits of a number.py
```diff
@@ -1,7 +1,13 @@
...
+    # The "q" value should be declared before using it(mandatory). And this value can be changed later.
+
+n=int(input("Enter Number: "))   # asking user for input
+while n>0:         # Until "n" is greater than 0, execute the loop. This means that until all the digits of "n" got extracted.
```
NitkarshChourasia, line 5: But buddy, I love your approach of understanding as well at the same time contributing to the repository.
It is like contributing + understanding the program, simultaneously.
I would too, try this approach.
But, it would be helpful, if you try to improve the execution and logic and readability of the program.

NitkarshChourasia (commented):
Good Work, though.

end
sarayusreeyadavpadala: Yeah! I've removed it.

Sum of digits of a number.py
sarayusreeyadavpadala: Yes! you said right. I'll do that today.

Sum of digits of a number.py
sarayusreeyadavpadala: Thanks for your suggestion NitkarshChourasia. I'll follow this approach.

NitkarshChourasia: @sarayusreeyadavpadala Do check my pull requests also.
Let's grow together.
I will always @sarayusreeyadavpadala assign you like this.
You too, try to assign me.
Because coding is not supposed to be lonely activity, but a fun and a social one.
That is called pair programming, I guess.
Right?
Let's do it.
I will post my programs.
See, what things I do.
Basically what I follow is:
I comment it.
Then, write down my basic requirements like,

"""
Requirements:
- DRY
- KISS
- OOP
- Function / Class
- snake_case
- Docstring 
- Comment
- unittest
- type hinting
- type checking
- Good logic (subjective)
- Readablility (subjective)
- Proper formatting using black
- Optimising the imports.

Etc...
This is what I like **AT LEAST** do...

Come on, all the best.

sarayusreeyadavpadala: #Let'sGrowTogether @NitkarshChourasia

sarayusreeyadavpadala: @sarayusreeyadavpadala review my code, I've changed it.

sarayusreeyadavpadala: @NitkarshChourasia view my code, I've changed it.

NitkarshChourasia: # Yes.

NitkarshChourasia (commented):
Good, but do improve the code further.

Use docstring to explain the code.
print(function_name.__doc__)

use
if __name__ == "__main__"
    function_name

NitkarshChourasia (commented):
Please improve it further.

PR #1966, opened 2023-08-12
code is changed

sarayusreeyadavpadala:
Hey geekcomputers! I have changed the code. Now, this code will ask the user to enter a number until the user enters the correct input. And performs addition after this. Let me know if any changes required.

### DIFF
--- a/VoiceRepeater/readme.md
+++ b/VoiceRepeater/readme.md
@@ -9,7 +9,3 @@
 ### And plays it!
 
 ### Requirements: Python, SpeechRecognition and playsound
-
-
-
-I am just trying to contribute please let me do this.....thank you :)


Sum of digits of a number.py
```diff
@@ -1,7 +1,13 @@
 q=0
-n=int(input("Enter Number: "))
-while(n>0):
- r=n%10
- q=q+r
- n=n//10
-print("Sum of digits is: "+str(q))
+while True:
```
NitkarshChourasia, line 2: Don't force the user to enter the number. 
Using while true loop.

NitkarshChourasia, line 2: Something like this would be better.
```
q = 0
k = 3
for i in range(k):
    n = input(f"Please enter a number({k - i}): ")
    if n.isnumeric():
        n = int(n)
        break
    else:
        print("enter integer only: ")
        continue
  ```
  k naming is for OOP implementation and to avoid **magic number**.
  Using continue and break is good practice.
  
  You can opt for the ({k -i}) formatting.
  It is just there to show user how many tries are left.

File: Sum of digits of a number.py [hunk 2]
```diff
@@ -1,7 +1,13 @@
...
+    n=input("enter a number:")
+    if n.isnumeric():
+        n=int(n)
+        break
+    else:
```
NitkarshChourasia, line 7: ### Encapsulate all of this in funciton please.

Sum of digits of a number.py
NitkarshChourasia: You can also use match case with "e", "exit", "exit()", "n", etc...for all neg values.
using isinstance of str checking, if yes then lower them, if that matches with these keywords then just break the loop.
Like this will enable exit the program.

NitkarshChourasia (changes requested):
# Please improve the code, now.

Sum of digits of a number.py
sarayusreeyadavpadala: Okay.

Sum of digits of a number.py
sarayusreeyadavpadala: Or can ask the user do you want to continue further when user entered wrong input in every iteration.

OfficialAhmed (approved):

Sum of digits of a number.py
NitkarshChourasia: Something like this would be better.
```
q = 0
k = 3
for i in range(k):
    n = input(f"Please enter a number({k - i}): ")
    if n.isnumeric():
        n = int(n)
        break
    else:
        print("enter integer only: ")
        continue
  ```
  k naming is for OOP implementation and to avoid **magic number**.
  Using continue and break is good practice.
  
  You can opt for the ({k -i}) formatting.
  It is just there to show user how many tries are left.

Comment (NitkarshChourasia) [reply Sum of digits of a number.py:2]: > Or can I ask the user do you want to continue further when user entered wrong input in every iteration.

You can also use match case with "e", "exit", "exit()", "n", etc...for all neg values.
using isinstance of str checking, if yes then lower them, if that matches with these keywords then just break the loop.
Like this will enable exit the program.

**Review by NitkarshChourasia — CHANGES_REQUESTED** [review]

File: Sum of digits of a number.py (hunk 1 above)
Comment (sarayusreeyadavpadala) [reply Sum of digits of a number.py:2]: > Don't force the user to enter the number. Using while true loop.

Or can I ask the user do you want to continue further when user entered wrong input in every iteration.

**Comment by sarayusreeyadavpadala** [comment]
import sys
q=0
def user():
    for i in range(3,0,-1):
        n=input("enter a number:")
        if n.isnumeric():
            n=int(n)
            return n
        else:
            print("enter integer only")
            print(f'{i-1} chances are left')
            continue
        
n=user()
if type(n) is type(None):
    print("Try again!")
    sys.exit()
while n>0:
    r=int(n%10)
    q=q+r
    n=n/10
print(q)

**Comment by sarayusreeyadavpadala** [comment]
import sys
q=0
def user():
    for i in range(3,0,-1):
        n=input("enter a number:")
        if n.isnumeric():
            n=int(n)
            return n
        else:
            print("enter integer only")
            print(f'{i-1} chances are left')
            continue
        
n=user()
if type(n) is type(None):
    print("Try again!")
    sys.exit()
while n>0:
    r=int(n%10)
    q=q+r
    n=n/10
print(q)

**Comment by sarayusreeyadavpadala** [comment]
`import sys
q=0
def user():
    for i in range(3,0,-1):
        n=input("enter a number:")
        if n.isnumeric():
            n=int(n)
            return n
        else:
            print("enter integer only")
            print(f'{i-1} chances are left')
            continue
        
n=user()
if type(n) is type(None):
    print("Try again!")
    sys.exit()
while n>0:
    r=int(n%10)
    q=q+r
    n=n/10
print(q)`

**Comment by sarayusreeyadavpadala** [comment]
```import sys
q=0
def user():
    for i in range(3,0,-1):
        n=input("enter a number:")
        if n.isnumeric():
            n=int(n)
            return n
        else:
            print("enter integer only")
            print(f'{i-1} chances are left')
            continue
        
n=user()
if type(n) is type(None):
    print("Try again!")
    sys.exit()
while n>0:
    r=int(n%10)
    q=q+r
    n=n/10
print(q)` ``

**Comment by sarayusreeyadavpadala** [comment]
how is this? @NitkarshChourasia

**Comment by NitkarshChourasia** [comment]
```python
import sys

def get_integer_input(prompt, attempts):
    for i in range(attempts, 0, -1):
        try:
            n = int(input(prompt))
            return n
        except ValueError:
            print("Enter an integer only")
            print(f"{i-1} {'chance' if i-1 == 1 else 'chances'} left")
    return None

def sum_of_digits(n):
    total = 0
    while n > 0:
        total += n % 10
        n //= 10
    return total

chances = 3
number = get_integer_input("Enter a number: ", chances)

if number is None:
    print("You've used all your chances.")
    sys.exit()

result = sum_of_digits(number)
print(f"The sum of the digits of {number} is: {result}")
```
These changes make the code more modular, easier to understand, and more user-friendly.

NitkarshChourasia:
use ```python.... to write code in .md format.... instead of plain ``` *2

sarayusreeyadavpadala:
The code I wrote, is easy to understand too. Here I didn't use exception handling. And beginner friendly too.

geekcomputers:
Can't be merged as there are conflicting files

NitkarshChourasia:
# Indeed.

NitkarshChourasia:
> Can't be merged as there are conflicting files

@geekcomputers # Indeed.

PR #1972, opened 2023-08-24
Create bodymass.py

erent8:
thanks for project.

### DIFF
--- a/bodymass.py
+++ b/bodymass.py
@@ -1,0 +1,19 @@
+kilo = float (input("kilonuzu giriniz(örnek: 84.9): "))
+boy = float (input("Boyunuzu m cinsinden giriniz: "))
+
+vki = (kilo / (boy**2))
+
+if vki < 18.5:
+    print(f"vucut kitle indeksiniz:  {vki} zayıfsınız.")
+elif vki < 25:
+    print (f"vucut kitle indeksiniz: {vki} normalsiniz.")
+elif vki < 30:
+    print (f"vucut kitle indeksiniz: {vki} fazla kilolusunuz.")
+elif vki < 35:
+    print (f"vucut kitle indeksiniz: {vki} 1. derece obezsiniz")
+elif vki < 40:
+    print (f"vucut kitle indeksiniz: {vki} 2.derece obezsiniz.")
+elif vki >40:
+    print (f"vucut kitle indeksiniz: {vki} 3.derece obezsiniz.")
+else:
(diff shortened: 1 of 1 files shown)


erent8:
thanks for project.

bodymass.py
```diff
@@ -0,0 +1,19 @@
+kilo = float (input("kilonuzu giriniz(örnek: 84.9): "))
```
NitkarshChourasia, line 1: Encapsulate it in function.

bodymass.py
```diff
@@ -0,0 +1,19 @@
...
+boy = float (input("Boyunuzu m cinsinden giriniz: "))
+
+vki = (kilo / (boy**2))
+
+if vki < 18.5:
+    print(f"vucut kitle indeksiniz:  {vki} zayıfsınız.")
+elif vki < 25:
+    print (f"vucut kitle indeksiniz: {vki} normalsiniz.")
```
NitkarshChourasia, line 9: If using so many if-elif-else statements, then use match-case, which is much more better.

bodymass.py
```diff
@@ -0,0 +1,19 @@
...
+elif vki < 30:
+    print (f"vucut kitle indeksiniz: {vki} fazla kilolusunuz.")
+elif vki < 35:
+    print (f"vucut kitle indeksiniz: {vki} 1. derece obezsiniz")
+elif vki < 40:
```
NitkarshChourasia, line 14: Don't use **Magic Numbers.**

bodymass.py
```diff
@@ -0,0 +1,19 @@
...
+    print (f"vucut kitle indeksiniz: {vki} 2.derece obezsiniz.")
+elif vki >40:
+    print (f"vucut kitle indeksiniz: {vki} 3.derece obezsiniz.")
+else:
+    print("Yanlış değer girdiniz.")
```
NitkarshChourasia, line 19: use
```python
if __name__ == "__main__":
        return main()
```

bodymass.py
```diff
@@ -0,0 +1,19 @@
+kilo = float (input("kilonuzu giriniz(örnek: 84.9): "))
+boy = float (input("Boyunuzu m cinsinden giriniz: "))
+
+vki = (kilo / (boy**2))
```
NitkarshChourasia, line 4: Please use a standard reading language.
# Please use **ENGLISH** language for writing programs with python, obviously.

NitkarshChourasia (changes requested):

PR #1974, opened 2023-08-28
Revert "Create bodymass.py"

NitkarshChourasia:
Reverts geekcomputers/Python#1972

Changes has not been resolved, and falsely marked **resolved.**


OfficialAhmed (changes requested):
Why removed bodymass.py ? 
You didnt change anything

NitkarshChourasia:
@OfficialAhmed Because he can understand English, and write and reads well in English, but was using a language in the content of the program, that was not comprehensible to everyone, so just told him to write it in English, so that everyone can understand it, because it is a repository available on a international level, and it will be a good practice and standard, if every program was in English, which are not a necessity or even need to be in any other language then English, but he was arrogant and being nationalist for no reason whatsoever, on such a small improvement I suggested.
Also, doing so will make the repository appear useless, because everyone will be using their language to write, which is not feasible and won't be a standard for the repository, I think.

NitkarshChourasia:
so, i just reverted, which is not merged, should have been, though.

PR #1975, opened 2023-09-02
Created power_of_n.py

HimanshuGupta-p1:
Created a program that calculates x raised to the power of n, where x can be a float number or integer and n can be a positive or negative integer.

Example 1:

Input: x = 2.00000, n = 10
Output: 1024.00000
Example 2:

Input: x = 2.10000, n = 3
Output: 9.26100
Example 3:

Input: x = 2.00000, n = -2
Output: 0.25000
Explanation: 2^-2 = 1/(2^2) = 1/4 = 0.25

### DIFF
--- a/power_of_n.py
+++ b/power_of_n.py
@@ -46,13 +46,12 @@ class Solution:
         return result
 
 
-if __name__ == "main":
-    obj = Solution() #Creating object of the class Solution
+obj = Solution() #Creating object of the class Solution
 
-    #Taking inouts from the user
-    x = float(input("Enter the base number: "))
-    n = int(input("Enter the power number: "))
+#Taking inouts from the user
+x = float(input("Enter the base number: "))
+n = int(input("Enter the power number: "))
 
-    #calling the function using object obj to calculate the power
-    answer = obj.binaryExponentiation(x, n)
-    print(answer) #answer
+#calling the function using object obj to calculate the power
+answer = obj.binaryExponentiation(x, n)
+print(answer) #answer
--- a/power_of_n.py
+++ b/power_of_n.py
@@ -1,57 +1,58 @@
-#Python program to calculate x raised to the power n (i.e., x^n)
+# Assign values to author and version.
+__author__ = "Himanshu Gupta"
+__version__ = "1.0.0"
+__date__ = "2023-09-03"
 
-# Script Name		: power_of_n.py
-# Author		    : Himanshu Gupta
-# Created			: 2nd September 2023
-# Last Modified		:
-# Version			: 1.0
-# Modifications		:
-# Description		: Program which calculates x raised to the power of n, where x can be float number or integer and n can be positive or negative number
-# Example 1:
+def binaryExponentiation(x: float, n: int) -> float:
+    """
+    Function to calculate x raised to the power n (i.e., x^n) where x is a float number and n is an integer and it will return float value
 
-# Input: x = 2.00000, n = 10
-# Output: 1024.00000
-# Example 2:
+    Example 1:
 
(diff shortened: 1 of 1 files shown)


power_of_n.py
```diff
@@ -0,0 +1,57 @@
...
+
+# Input: x = 2.10000, n = 3
+# Output: 9.26100
+# Example 3:
+
+# Input: x = 2.00000, n = -2
+# Output: 0.25000
+# Explanation: 2^-2 = 1/(2^2) = 1/4 = 0.25
+
+#Class
```
NitkarshChourasia, line 24: Leave a space after '#'.

power_of_n.py
```diff
@@ -0,0 +1,57 @@
+#Python program to calculate x raised to the power n (i.e., x^n)
+
+# Script Name		: power_of_n.py
```
NitkarshChourasia, line 3: Use """ docstring """ instead of comments.

power_of_n.py
```diff
@@ -0,0 +1,57 @@
...
+# Author		    : Himanshu Gupta
```
NitkarshChourasia, line 4:
```python
# use 
__author__ == "Himanshu Gupta"

# try printing it.

print(__author__)
```

power_of_n.py
```diff
@@ -0,0 +1,57 @@
...
+# Output: 9.26100
+# Example 3:
+
+# Input: x = 2.00000, n = -2
+# Output: 0.25000
+# Explanation: 2^-2 = 1/(2^2) = 1/4 = 0.25
+
+#Class 
+class Solution:
+
```
NitkarshChourasia, line 26: Also using docstring in class like this: to explain the brief task the program does, would be helpful.

```python
""" To output a given number to the power of n """

# if this is used for printing, see what happens.
# This is inbuilt dunder method.


print(function_name.__docs__)
```

power_of_n.py
```diff
@@ -0,0 +1,57 @@
...
+        return result
+
+
+obj = Solution() #Creating object of the class Solution
+
+#Taking inouts from the user
+x = float(input("Enter the base number: "))
+n = int(input("Enter the power number: "))
+
+#calling the function using object obj to calculate the power
```
NitkarshChourasia, line 55:
```python
# use this for executing the program when ran directly.

if __name__ == "__main__":
        print(function_name(param1, param2))
        
```

NitkarshChourasia: The equivalent is javascript would be...

```javascript

if (require.main == module) {
    console.log(function_name(param1))
    
```

power_of_n.py
```diff
@@ -0,0 +1,57 @@
...
+answer = obj.binaryExponentiation(x, n)
+print(answer) #answer
```
NitkarshChourasia, line 57: # PLEASE DO ALL THIS CHANGES BEFORE MERGING, RESPECTFULLY.

NitkarshChourasia (changes requested):

power_of_n.py
HimanshuGupta-p1: Okay !.

power_of_n.py
```diff
@@ -0,0 +1,58 @@
...
+    Example 1:
+
+    Input: x = 2.00000, n = 10
+    Output: 1024.0
+    Example 2:
+
+    Input: x = 2.10000, n = 3
+    Output: 9.261000000000001
+
+    Example 3:
```
NitkarshChourasia, line 19: All perfect!

NitkarshChourasia (commented):
All perfectly done.

PR #1988, opened 2023-10-03
Create TowerOfHanoi.py

codingCapricorn:
Tower of Hanoi using Python

### DIFF
--- a/TowerOfHanoi.py
+++ b/TowerOfHanoi.py
@@ -1,0 +1,14 @@
+# Recursive Python function to solve the tower of hanoi --
+
+def TowerOfHanoi(n , source, destination, auxiliary):
+	if n==1:
+		print ("Move disk 1 from source",source,"to destination",destination)
+		return
+	TowerOfHanoi(n-1, source, auxiliary, destination)
+	print ("Move disk",n,"from source",source,"to destination",destination)
+	TowerOfHanoi(n-1, auxiliary, destination, source)
+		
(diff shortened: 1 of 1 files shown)


TowerOfHanoi.py
```diff
@@ -0,0 +1,14 @@
...
+def TowerOfHanoi(n , source, destination, auxiliary):
+	if n==1:
+		print ("Move disk 1 from source",source,"to destination",destination)
+		return
+	TowerOfHanoi(n-1, source, auxiliary, destination)
+	print ("Move disk",n,"from source",source,"to destination",destination)
+	TowerOfHanoi(n-1, auxiliary, destination, source)
+		
+# Driver code
+n = 4
```
NitkarshChourasia, line 12: Use 
```python
if __name__ == "__main__":
    # Function to execute.

File: TowerOfHanoi.py [hunk 2]
```diff
@@ -0,0 +1,14 @@
+# Recursive Python function to solve the tower of hanoi --
+
```
NitkarshChourasia, line 2: Short and simple.

NitkarshChourasia (approved):
Better if implemented some more things.

TowerOfHanoi.py
rein14: this is good practice

NitkarshChourasia: Yes. Indeed.

Issue #1910, opened 2023-06-06
Using elif instead of if and nested else-if


hysteri4:
Hello, 

I'm nowhere near to be good neither at Github nor at Python. But in your Ceaser decipher script, I saw that you used nested `if-else if` blocks instead of using the more suitable `if-elif` conditions. I thought that would be better both for you and for the readability of the code. Whatever you choose to do so, I wish you a good day.

deepteshgurav:
Hi. Can I work on this issue? I believe we can optimize the nested conditions using elif statement. 

"""
if func == 2:
         decode()
elif func == 1:
        encode()
else:
        print("\n[>] Invalid input")
        exit()
"""

Using this we can optimize and improve the code readability.

PR #1992, opened 2023-10-05
FIXED: Caesar cipher decoder

OfficialAhmed:
* [**FIXED  ISSUE**] #1910 
* [**ENHANCED SECURITY**] The key is no longer limited to 25 can be any number, hence, brute forcing takes longer to decrypt
  * The correct decoded cipher text is returned. No longer returns the possible solutions like before.
    - `Encode` The key is added to the ciphered text 
    - `Decode` The key is removed from the ciphered text 
* [**ENHANCED ERROR HANDLING**] Wrong input no longer terminates the script. It'll rerun for the correct input


NitkarshChourasia (approved):
Perfect!

Issue #1573, opened 2022-06-28
How do I create a generator 


La-victoire:

kumarvivek1752:
simply generator is a function but we write yield instead of return because when we write return in our function the function end on return. however, any of the local variables created during the function scope are destroyed and we can not accessible further. While in case of generator when it encounters a yield keyword the state of the function is frozen and all the variables are stored in memory until the generator is called again

basic syntax
```

def simpleGeneratorFun():
    yield 1
    yield 2
    yield 3
   
# x is a generator object
x = simpleGeneratorFun()
  
# Iterating over the generator object using next
print(x.next()) # In Python 3, __next__()
print(x.next())
print(x.next())
```

Aliwahid17:
yes, you tell right about generators but instead of writing print statements multiple times use for or while loop

```python
def simpleGeneratorFun():
    yield 1
    yield 2
    yield 3
   
# x is a generator object
x = simpleGeneratorFun()
  
# Iterating over the generator object using next
[-]print(x.next()) # In Python 3, __next__()
[-]print(x.next())
[-]print(x.next())

[+]for i in range(3):
 [+]   print(x.next())
```

Aliwahid17:
```
+ ji
```

kumarvivek1752:
> yes, you tell right about generators but instead of writing print statements multiple times use for or while loop
> 
> ```python
> def simpleGeneratorFun():
>     yield 1
>     yield 2
>     yield 3
>    
> # x is a generator object
> x = simpleGeneratorFun()
>   
> # Iterating over the generator object using next
> print(x.next()) # In Python 3, __next__()
> print(x.next())
> print(x.next())
> 
>  for i in range(3):
>      print(x.next())
> ```

yeah but that is only for demonstration not for realworld use.

Aliwahid17:
oh ok. by the way good demonstration

Tevazu:
#### _Generators are objects used to create an iterable object in Python and do not take up memory space. For example, if we generate 100000 values and keep these values in memory, it will take up a lot of space, but this problem can be overcome thanks to generators. You can check this simple program which takes squares of the numbers 1, 2, 3, 4, 5 :_

```
def take_square():
    for i in range(1,6):
        yield (i ** 2)
```
#### _You can compare with the version which we didn't use generator:_
```
def take_square():
    result = []
    for i in range(1,6):
        result.append(i ** 2)
    return result
```

Aliwahid17:
[...]

PR #2002, opened 2023-10-11
generator.py

MohitHire:
#1573


NitkarshChourasia (changes requested):
Simple practice programs are not accepted.
Sorry.

PR #2014, opened 2023-10-21
Added English text to pig Latin converter

usmansafdarktk:


NitkarshChourasia (changes requested):
Adding a README.md on how to use it, etc... would be beneficial.

PR #2018, opened 2023-10-24
Create Roman number to decimal

collateralusr:
The value function is used to map Roman numerals to their decimal values, and the romanToDecimal function converts a Roman numeral to an integer using the given mapping.

### DIFF
--- a/Roman number to decimal
+++ b/Roman number to decimal
@@ -1,0 +1,40 @@
+
+def value(r):
+    if (r == 'I'):
+        return 1
+    if (r == 'V'):
+        return 5
+    if (r == 'X'):
+        return 10
+    if (r == 'L'):
+        return 50
+    if (r == 'C'):
+        return 100
+    if (r == 'D'):
+        return 500
+    if (r == 'M'):
+        return 1000
+    return -1
+ 
+def romanToDecimal(str):
+    res = 0
+    i = 0
+ 
+    while (i < len(str)):
+        s1 = value(str[i])
+        if (i + 1 < len(str)):
+            s2 = value(str[i + 1])
+            if (s1 >= s2):
+                 res = res + s1
+                 i = i + 1
+            else:
+                res = res + s2 - s1
+                i = i + 2
+        else:
+            res = res + s1
+            i = i + 1
+ 
+    return res
+
+romannumber = input("Enter the Roman Number you want to convert: ")
+print("Integer form of Roman Numeral is: " + str(romanToDecimal(romannumber)))


vanshdhiwar09:
`nterms = int(input("How many numbers"))
n1,n2 = 0,1
count = 0
if nterms <= 0:
    print("Please enter a positive integer")
elif nterms == 1:
    print("Fibonnaci sequence upto",nterms,":")
    print("n1")
else:
    print("Fibonnaci sequence.")
    while count < nterms:
        print(n1)
        nth = n1 + n2
        n1 = n2
        n2 = nth
        count += 1`

Roman number to decimal
```diff
@@ -0,0 +1,40 @@
+
+def value(r):
```
NitkarshChourasia, line 2: Use switch statement, they are far better instead of this.

Roman number to decimal
```diff
@@ -0,0 +1,40 @@
...
+    return -1
+ 
+def romanToDecimal(str):
+    res = 0
+    i = 0
+ 
+    while (i < len(str)):
+        s1 = value(str[i])
+        if (i + 1 < len(str)):
+            s2 = value(str[i + 1])
```
NitkarshChourasia, line 26: leetcode problem idea...I see.

NitkarshChourasia (approved):

dexterharrison29 (commented):
this code is so fucking awful... Lemme pullup with a better one

Roman number to decimal
dexterharrison29: Yea

Roman number to decimal
dexterharrison29, line 1: heres a version of this designed to be more efficient
`def value(r):
    sym = ['M', 'CM', 'D', 'CD', 'C', 'XC', 'L', 'XL', 'X', 'IX', 'V', 'IV', 'I']  
    num = [1000, 900, 500, 400, 100, 90, 50, 40, 10, 9, 5, 4, 1]  
    result = 0  
    for idx, val in enumerate(r):  
        first_num = num[sym.index(val)]  
        second_num = num[sym.index(r[idx + 1])] if idx + 1 != len(r) else -1  
        if first_num >= second_num:  
            result += first_num  
        else:
            result -= first_num
    return result`

dexterharrison29 (commented):
I made this but more efficient here. https://github.com/dexterharrison29/code-review/blob/main/roman-numerals.py

AbhyudayMeher2005 (commented):
Good

PR #2021, opened 2023-10-24
calculator.py minor improvement

ilia-abbasi:
Prioritized lowering the input for edge cases like "Mod" and "Rad"


NitkarshChourasia (changes requested):
Unnecessary change.

PR #2029, opened 2023-11-01
Added Industrial_developed_hangman

danila12315454:
Added example of Industrial developed game Hangman


NitkarshChourasia (changes requested):
A screenshot would be better if it is embedded in to README.md file.

danila12315454:
screenshot of programm working?

NitkarshChourasia:
> screenshot of programm working?

Yes, would be better. This is all perfect. Just for those who wanna just scroll through the repo so that they can know better, what is this.

danila12315454:
> A screenshot would be better if it is embedded in to README.md file.

done

danila12315454:
I also have a question about adding github workflow in file. Will it work correctly? I am talking about yml file

NitkarshChourasia:
@OfficialAhmed Please, check this out.

OfficialAhmed (approved):
All good however
- **_Scraping a word_**: I dont recommend scraping a web page for a word. I rather use an api to fetch words which is much safer and better approach to this problem. Please consider the policy and the rules of the website that is being scraped it might be illegal to do so or they dont allow that. 

- **_Hard coded_**: It would be better to first check if the HTML element that contains the word exist before scraping because this could be an issue the future if the website gets changed or updated

PR #2050, opened 2023-12-09
Update Python Program to Sort Words in Alphabetic Order.py

AndrewB50:
The .sort() function sorts strings by ASCII Code. The original code did not return all words alphabetically. I added a .lower() function to make sure all words were the same case that way they would return alphabetically.

### DIFF
--- a/Python Program to Sort Words in Alphabetic Order.py
+++ b/Python Program to Sort Words in Alphabetic Order.py
@@ -1,18 +1,18 @@
-# Program to sort alphabetically the words form a string provided by the user
+# Program to sort words alphabetically and print them out in a list
 
+#declaring variables
 my_str = "Hello this Is an Example With cased letters"
+word_Dict = {}
+count = 0
+
+#Need to make all words the same case, otherwise, the .sort() function sorts them by ASCII code and they will not appear alphabetically. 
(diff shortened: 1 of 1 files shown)


Python Program to Sort Words in Alphabetic Order.py
```diff
@@ -1,18 +1,18 @@
-# Program to sort alphabetically the words form a string provided by the user
+# Program to sort words alphabetically and print them out in a list
 
+#declaring variables
 my_str = "Hello this Is an Example With cased letters"
+word_Dict = {}
```
NitkarshChourasia, line 5: Why this dictionary is used for no reason at all?

Python Program to Sort Words in Alphabetic Order.py
```diff
@@ -1,18 +1,18 @@
...
+count = 0
```
NitkarshChourasia, line 6: Why this variable **count** is used for no reason of it being anywhere in the program.

NitkarshChourasia (changes requested):
It is good. 
But can be made much better.
By making sure the cases of the letters stays consistent, and not every letter become single case.

Best regards,
@NitkarshChourasia

PR #2056, opened 2023-12-14
this changes files extension

devpoisons:
password is 4121
enter the names of files or use these two wildcards
* selects all r give random files range from 1 to 5 files 

enter the extensions add a dot before any extension e.g .mp4, .mp3 you can give more than one extension and it wil arrange them according to the files. Good luck. 

email: devpoison@outlook.com


devpoisons:
adding to the python programs

PR #2079, opened 2024-01-07
add: gtts functional class.

NitkarshChourasia:
A program with features packed to work with Google's Speech API, respectfully.


NitkarshChourasia:
@PBJI

PR #2085, opened 2024-01-09
edited input line of code

SkJishan04:
Without taking integer input in a variable I have directly added input code in the function . This will reduce the space .

### DIFF
--- a/Patterns/pattern1.py
+++ b/Patterns/pattern1.py
@@ -7,8 +7,7 @@
 #1
 
 def main():
-    lines = int(input("Enter no.of lines: "))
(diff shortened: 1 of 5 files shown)


NitkarshChourasia (approved):
Good work.
But I guess. 
The paradigm of OOP is to be maintained. 
Right?

So, therefore we use that kind of approach.
Still this is fine.
But maybe considered irrelevant as the data stored in the variable is not a very big data that consumes memory.

Best regards,
@NitkarshChourasia 


@geekcomputers Please merge this Pull Request.


Best regards,
@NitkarshChourasia

PR #2171, opened 2024-05-04
Update main.py

nadirhsan:
changed the y or n input to a loop to make sure program works even with wrong input

ps: idk what i'm doing , this is my first time on github


NitkarshChourasia (commented):
You implemented a system.
Which makes sure if any other input is entered other then the required.
It will again ask for the input.
Thanks for your contributions.

Best regards,
@NitkarshChourasia

PR #2191, opened 2024-05-13
refactor: Improve security scripts

NitkarshChourasia:
refactor: Improve security scripts already present.


NitkarshChourasia:
Thanks @geekcomputers

PR #2201, opened 2024-05-18
I checked the repository using black(plugin)

Brevno78:


NitkarshChourasia (changes requested):
No mass editing should be done. Rejected! Close this Pull Request. If owner @geekcomputers  closes this, then it is not good for the long term @Brevno78 . Thank you.

Warm regards,
@NitkarshChourasia

PR #2245, opened 2024-06-18
Update tictactoe.py

shritej1808:
new pull


NitkarshChourasia (approved):
Thanks for your contribution.
@shritej1808 
Do mind also removing the whitespaces as those are left in it.

Thanks for your contribution.
@NitkarshChourasia

NitkarshChourasia (commented):
Good.

PR #2251, opened 2024-06-20
[issue:#2010]:python file to add_two_number

Anxhul10:
This file is able to take two int values and print their sum .
Visual representation of program is given below:-
Screenshot 2024-06-20 101933


NitkarshChourasia (commented):
Please don't submit basic problems.

PR #2398, opened 2024-10-26
Change import statements to use only used classes

Ferszus:
Updated import statements to avoid from module import * usage. Importing specific functions or classes improves code readability, minimizes the risk of name conflicts, and enhances performance by only loading necessary components. This approach also facilitates better static analysis and autocompletion support.


NitkarshChourasia (commented):
Good.

PR #2529, opened 2025-02-16
 8_puzzle.py

ayushh0406:
This code defines a PuzzleState class to represent each state of the puzzle. It uses the A* search algorithm to find the shortest path from the initial state to the goal state. The Manhattan distance heuristic is used to guide the search. The solve_puzzle function returns the solution state, and the print_solution function prints the sequence of moves to solve the puzzle.


ayushh0406:
Implement Eight Puzzle Solver using A Algorithm with Manhattan Heuristic

PR #2555, opened 2025-03-07
Fix assembler.py

parsajavanshir:
Fix: Bug fixes in stack management, division by zero, and string handling

- Fixed stack underflow in `pop` command
- Handled division by zero in `div` instruction
- Validated empty files before processing
- Added exception handling for return stack
- Improved string assignment in `mov` instruction


pierzoe:
unsubscribe

PR #2785, opened 2025-07-15
ruff fixes

lighting9999:
This is Updates file to the ruff fixes. ruff fixes 136 errors and edit this 73 files.


lighting9999:
@geekcomputers But this pull request it seem like  readme is wrong.

Issue #2843, opened 2025-08-17
Collatz Sequence


FROGROWL:
Creating a program to let the user see the process of in a collatz sequence of their inputs. Just a simple programming : )

ani-devsx998:
Hello everyone, I am a newbie to open source, can anyone guide me what is this issue about?

PR #2845, opened 2025-08-18
Step-By-Step Collatz Sequence #2843

BhavanKumarGM:
🔹 Collatz Sequence Problem

The Collatz sequence (also called the 3n + 1 problem) is a simple mathematical process:
           1. Start with any positive integer n.
           2. If n is even, divide it by 2 → n = n / 2.
           3. If n is odd, multiply it by 3 and add 1 → n = 3n + 1.
           4. Repeat this process with the new value of n.
           5. Eventually (for all tested numbers), the sequence will reach 1.
          
Explaining the code part:
    **Description**  
              When the user enters a non-integer (like "abc"), the program raises a ValueError and exits immediately. It would be better if the program handled this by showing an error message and requesting valid input again.

    **Steps to reproduce**  
              1. Run the program (e.g., in Python).  
              2. Input “abc” instead of a number.

     **Expected behavior**  
              Program prints something like “Please enter a valid positive integer.” and then prompts again.

     **Actual behavior**  
              Program crashes with a traceback, e.g., `ValueError: invalid literal for int() with base 10: 'abc'`.

     **Suggested fix**  
              Wrap the input logic in a loop with `try/except`, and use `continue` in the except block to reprompt.

     **The Main Function**
                 while True:
                           try:
                                num = int(input("Enter a positive integer: "))
                                 if num > 0:
                                       break
                                 else:
                                      print("Please enter a positive integer.")
                           except ValueError:
                                     print("Invalid input! Please enter a number.")
                                     continue
     **Conclusion** 
                 This approach ensures the program doesn't crash on bad input and prompts the user again in a friendly way.

### DIFF
--- a/Collatz Sequence/Collatz Sequence.py
+++ b/Collatz Sequence/Collatz Sequence.py
@@ -1,29 +1,23 @@
-def collatz_steps(n):
-    times = 0
+def collatz_sequence(n):
+    """Generate and print the Collatz sequence for n."""
+    steps = [n]
     while n != 1:
         if n % 2 == 0:
-            print(f"{n} / 2", end=" ")
-            n = n // 2 # "//" is a floor division where it rounds down the result
+            n = n // 2
         else:
-            print(f"{n} * 3 + 1", end=" ")
             n = 3 * n + 1
-        print(f"= {n}")
-        times += 1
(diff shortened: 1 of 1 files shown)


BhavanKumarGM:
Kindly review this and let me know if any modifications are required.

